Issue #37, opened 2026-05-01
[DOC] Documentation drift: tutorials reference non-existent binaries and stale CMake defaults


RamyaGuru:
## Summary

  Several documentation files were authored against an earlier code shape that has since been refactored on `main` (raw/socket bench split, socket backend addition). The
   references were never updated, so the published tutorials walk readers through commands that don't run and CMake settings that don't match the source. Filing this as
  a single issue rather than per-file because the root cause is shared and the fix should be paired with automated drift detection.

  ## Stale references

  ### `docs/tutorials/benchmarking_examples.md`

  - **L7, L144, L152** — references the binary `daqiri_bench_default` and the config `daqiri_bench_default_tx_rx.yaml`. Neither exists on `main`. Actual binaries:
  `daqiri_bench_raw_gpudirect`, `daqiri_bench_raw_hds`, `daqiri_bench_raw_reorder_seq`, `daqiri_bench_raw_reorder_quantize`, `daqiri_bench_rdma`, `daqiri_bench_socket`
  (see `examples/CMakeLists.txt:59-71`). Suggested replacement for the loopback walkthrough: `daqiri_bench_raw_gpudirect` with `daqiri_bench_raw_sw_loopback.yaml` (no
  physical link required).
  - Internal cross-link `../getting-started.md#building-from-source` — anchor no longer exists; the section was renamed to "Build the DAQIRI Library" during the rewrite.

  ### `docs/getting-started.md`

  - The CMake options table lists `DAQIRI_MGR` default as `"dpdk rdma"`. Actual default per `src/CMakeLists.txt:137` is `"dpdk socket"` (which auto-pulls `rdma` due to
  the dependency rule in `src/CMakeLists.txt:144-152`).
  - Same table's "Valid values" line omits `socket`. Should be `dpdk`, `socket`, `rdma`.
  - The CMake build example uses `-DDAQIRI_MGR="dpdk rdma"`. Should be `"dpdk socket rdma"` (or just `"dpdk socket"` to match the documented default).
  - "Next Steps" → "Benchmarking Examples" bullet says "run `daqiri_bench_default` with a loopback test." Same as benchmarking_examples.md.
  - `../Dockerfile` link does not resolve under MkDocs. Use the absolute GitHub URL (`https://github.com/NVIDIA/daqiri/blob/main/Dockerfile`) for parity with the
  bare-metal CMake tab, which already does this.

  ### `docs/api-guide.md`

[...]

PR #40, opened 2026-05-01
#37 - Add CI gate for documentation drift

RamyaGuru:
Adds a check job that runs on every PR touching docs/, mkdocs.yml, the workflow, or the link-checker script, and restructures the deploy job to gate on the check passing. Two safeguards:

1. `mkdocs build --strict` with an explicit `validation:` block in mkdocs.yml that escalates link, anchor, and nav warnings (broken internal markdown links, missing nav files, broken anchors, absolute paths) from INFO to WARNING. `--strict` then fails the build on any of them. Without this block, MkDocs treats those as informational and `--strict` lets them through.

2. `scripts/check_html_links.py` parses every `href` in the hand-maintained `docs/index.html` and `docs/daqiri-api.html`, resolves each against the rendered `site/` directory, and fails if any target is missing. MkDocs' link graph does not cover these standalone HTML pages, so the script closes that gap.

Together these catch the class of drift tracked in #37: stale binary names in markdown, broken cross-links after a doc reorg, and unresolvable hrefs on the landing page after MkDocs URL changes.

Also drops the obsolete unified_docs / website / tutorial branches from the workflow trigger list — only main triggers a deploy now.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -10,6 +10,10 @@ on:
       - 'mkdocs.yml'
       - '.github/workflows/docs.yml'
       - 'scripts/check_html_links.py'
+      - 'scripts/check_doc_refs.py'
+      - 'examples/CMakeLists.txt'
+      - 'examples/*.yaml'
+      - 'README.md'
 
 permissions:
   contents: write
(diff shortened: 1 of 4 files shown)


RamyaGuru:
CI would now fail on broken markdown links/anchors/nav entries, dead hrefs in the hand-maintained landing & API HTML pages, and stale daqiri_bench_* binary or *.yaml config names referenced in docs that don't match examples/.

PR #48, opened 2026-05-05
#44 - Auto-bump num_bufs below 1.5x DPDK ring size with a warning

### LINKED ISSUE #44: [BUG] TX/RX bench stops after sending 14,336 packets on DGX Spark
**Describe the bug**

  daqiri_bench_raw_gpudirect over the QSFP loopback on DGX Spark transmits a fixed 14,336 packets (= 14 × batch_size=1024) regardless of how long the bench runs and regardless of whether memory_regions[*].kind is "huge" or "host_pinned". After those 14 bursts the bench keeps spinning but no further traffic is generated. RX matches TX exactly, so this is a TX-side stall, not loss. Effective throughput collapses with longer runs (~22 MB/s at --seconds 5, ~3.7 MB/s at --seconds 30 — the same 110 MB total in both cases) — orders of magnitude below ConnectX-7 line rate.

  YAML (Spark PCIe addresses inlined; the kind value is the only knob varied):

  ```
%YAML 1.2
  ---
  daqiri:
    cfg:
      version: 1
      stream_type: "raw"
      master_core: 16
      log_level: "info"
      loopback: ""
      memory_regions:
      - { name: "Data_TX_GPU", kind: "huge", affinity: 0, num_bufs: 8192, buf_size: 8064 }
      - { name: "Data_RX_GPU", kind: "huge", affinity: 0, num_bufs: 8192, buf_size: 8064 }
      interfaces:
      - name: "tx_port"
        address: 0000:01:00.0
        tx:
          queues:
          - { name: "tx_q_0", id: 0, batch_size: 1024, cpu_core: 17, memory_regions: ["Data_TX_GPU"], offloads: ["tx_eth_src"] }
      - name: "rx_port"
        address: 0002:01:00.0
        rx:
          flow_isolation: true
          queues:
          - { name: "rq_q_0", id: 0, cpu_core: 18, batch_size: 1024, memory_regions: ["Data_RX_GPU"] }
          flows:
          - { name: "flow_0", id: 0, action: { type: queue, id: 0 }, match: { udp_src: 4096, udp_dst: 4096 } }
  bench_rx:
    interface_name: "rx_port"
  bench_tx:
    interface_name: "tx_port"
    batch_size: 1024
    payload_size: 8000
    header_size: 64
    eth_dst_addr: <RX_MAC>
    ip_src_addr: 1.1.1.1
    ip_dst_addr: 2.2.2.2
    udp_src_port: 4096
    udp_dst_port: 4096
```

  Run the same bench at two different durations:

 ```
 docker run --rm --privileged --gpus all --network host \
    -v /mnt/huge:/mnt/huge \
    -v /tmp/cfg.yaml:/tmp/cfg.yaml:ro \
    daqiri:local \
    /opt/daqiri/bin/daqiri_bench_raw_gpudirect /tmp/cfg.yaml --seconds 5
```

  ```
docker run --rm --privileged --gpus all --network host \
    -v /mnt/huge:/mnt/huge \
    -v /tmp/cfg.yaml:/tmp/cfg.yaml:ro \
    daqiri:local \
    /opt/daqiri/bin/daqiri_bench_raw_gpudirect /tmp/cfg.yaml --seconds 30
```

  Tail of both runs (identical):

```
  [INFO] daqiri_dpdk_mgr.cpp:3814: Total packets transmitted by application (port/queue 0/0): 14336
  [INFO] daqiri_dpdk_mgr.cpp:3676: Total packets received by application (port/queue 1/0): 14336
  RX complete: packets=14336 bytes=115605504 bursts=14
```

The same exact 14336 / 14 numbers fall out for --seconds 5, --seconds 30, with kind: "huge", and with kind: "host_pinned" (after PR #41). Switching batch_size produces a proportional change (it's always exactly 14 bursts of batch_size), so the cap is in burst count, not in packet count.

  Other extended-stat noise (consistent across runs):
  - rx_q0_errors: ~640M–990M and rx_mbuf_allocation_errors: ~640M–990M — the RX queue scales with run duration here, so the RX core is alive and the NIC is back-pressuring against an empty mempool while TX is idle.
  - tx_phy_packets / tx_unicast_packets: 14336 — the NIC really did transmit 14336 unicast frames; the bench just doesn't submit   any more.

  **Expected behavior**

  Total transmitted-packets count should grow with --seconds, and effective bandwidth should approach line rate (or at least the configured pool/queue caps). On a 200 Gbps ConnectX-7 with 8 KB payloads, even a 10× safety margin would put us in the millions-of-packets-per-second range, not 14336 total per run.

  - Environment location: Bare-metal host (Spark), bench inside Docker.
  - Method of DAQIRI install: source build via scripts/build-container.sh (BASE_TARGET=dpdk DAQIRI_MGR="dpdk rdma socket").
  - Run command: as above (docker run --rm --privileged --gpus all --network host -v /mnt/huge:/mnt/huge -v
  /tmp/cfg.yaml:/tmp/cfg.yaml:ro daqiri:local …).

  **Environment details**

  - OS and version: Ubuntu 24.04 LTS host (kernel 6.17.0-1014-nvidia, aarch64); container based on
  nvcr.io/nvidia/cuda:13.1.0-devel-ubuntu24.04.
  - Python version: n/a (bench is the C++ daqiri_bench_raw_gpudirect).
  - DAQIRI version: reproduced on main at eb9c6d6 with kind: "huge", and on PR #41 head (72294fe) with both kind: "huge" and
  kind: "host_pinned".
  - Hardware configuration: DGX Spark — GB10 integrated GPU (sm_121, INTEGRATED=1), ConnectX-7 (TX 0000:01:00.0 / mlx5_0, RX
  0002:01:00.0 / mlx5_2, single QSFP cable looping the two chips). 4 KiB host pages, 3×1 GiB hugepages mounted at /mnt/huge.
  - Other relevant package versions: NVIDIA driver 580.142, CUDA 13.1, DPDK 25.11 with vendored dmabuf.patch and dpdk.nvidia.patch applied, libibverbs/mlx5 from DOCA 3.2.1 repo. isolcpus=16-19, nohz_full=16-19, rcu_nocbs=16-19, MRRS 4096 on all four CX-7 PFs (per Holoscan high-perf-networking tuning).

### PR DESCRIPTION
**Pull request by dleshchev**
## Summary

Closes #44 (the bench-stalls-after-14336-packets symptom).

The bench's TX/RX worker silently stalls when a memory region's `num_bufs` barely covers — or matches — the NIC descriptor ring size. The total packet count is then a fixed number regardless of `--seconds`. The existing guard at `daqiri_dpdk_mgr.cpp:1878` (RX) and `:1969` (TX) used `num_bufs < default_num_*_desc` (strict less-than), which lets equality and slightly-over configurations slip through to the deadlock.

Per Cliff's review, the better fix is to recover from the misconfig rather than refuse it: this PR detects any MR with `num_bufs < 1.5 × ring` in `adjust_memory_regions()` (which runs before `allocate_memory_regions()`, so the underlying GPU/CPU buffer is sized correctly) and bumps it to `3 × ring` (= 24576 with the default 8192 descriptors), emitting a `WARN` that names the MR, the requested value, the threshold, and the adjusted value. The 1.5× empirical deadlock boundary itself is unchanged.

## Investigation

Reproduced #44 on DGX Spark + ConnectX-7 (ring = 8192). Swept `num_bufs` while holding `batch_size = 1024` and the rest of the config constant:

| `num_bufs` | ratio to ring | Pre-fix result | App TX in 5 s |
|---|---|---|---|
| 8192 | 1.00× | stall | 14,336 |
| 9011 | 1.10× | stall | 15,360 |
| 10240 | 1.25× | stall | 7,627,776 (intermittently fine, but flaky) |
| 12288 | 1.50× | line rate | 7,591,936 |
| 16384 | 2.00× | line rate | 7,628,800 |

The threshold for reliable line-rate operation is **≥ 1.5× ring size**. Below that the bench silently stalls, producing duration-independent throughput numbers that look like a perf bug but are a config artifact (most likely TX/RX-pool exhaustion: every buffer ends up in the NIC's ring with no replacements, so the next `rte_pktmbuf_alloc` blocks forever).

## Verification (post-fix)

Re-running the four-point sweep on IGX + ConnectX-7, ring = 8192, `batch_size = 1024`, split-bench (TX and RX in separate containers):

| `num_bufs` | ratio | Behavior | TX in 5 s | RX in 5 s |
|---|---|---|---|---|
| 8192 | 1.00× | `WARN` + auto-bump to 24576 → runs | 7,342,592 | 7,342,592 |
| 10240 | 1.25× | `WARN` + auto-bump to 24576 → runs | 7,346,944 | 7,346,944 |
| 12288 | 1.50× | runs unchanged | 7,541,760 | 7,541,760 |
| 16384 | 2.00× | runs unchanged | 7,479,296 | 7,479,296 |

TX==RX exactly across all four cases. The previously-deadlocked configs now run at line rate (~95 Gbps) on the bumped pool. The `WARN` text:

```
[WARN] MR 'Data_TX_GPU' had num_bufs=8192 which is below the 12288 threshold
       (1.5x the 8192 NIC descriptors) and would deadlock the worker once the
       ring fills. Bumping to 24576 (3x ring).
```

## Caveats

- The check uses `default_num_rx_desc` / `default_num_tx_desc` (the **requested** descriptor count, before `rte_eth_dev_adjust_nb_rx_tx_desc` clamps to driver capabilities at line 2098). On NICs that clamp the ring *down*, the bump would over-allocate. For mlx5 / ConnectX-7 + 8192 this is a no-op so it doesn't matter today, but if precision matters later the check should move after `adjust_nb_rx_tx_desc`.
- 1.5× is the empirical deadlock boundary; 3× is the bumped-to value (so users get safe headroom, not just the bare minimum). Both are picked from the sweep above.
- The single check in `adjust_memory_regions()` runs once per MR rather than once per RX-queue + once per TX-queue, so an MR shared between RX and TX is bumped exactly once.

## Test plan

- [x] Builds cleanly inside the standard `scripts/build-container.sh` flow (`BASE_TARGET=dpdk DAQIRI_MGR="dpdk rdma socket"`).
- [x] All four `num_bufs` ratios produce the documented behavior on IGX + ConnectX-7. DGX Spark and IGX are the two platforms supported today; x86 + dGPU coverage is deferred.
- [x] Documented the new auto-bump behavior in `docs/configuration.md` and `docs/tutorials/configuration-walkthrough.md` (and unified the rule-of-thumb between them: the reference said "3x `batch_size`" while the walkthrough said "5x"; both now say "3x-5x").

### CODE REVIEW

File: src/managers/dpdk/daqiri_dpdk_mgr.cpp [hunk 1]
```diff
@@ -1875,9 +1875,12 @@ void DpdkMgr::initialize() {
         const auto& mr = cfg_.mrs_[q.common_.mrs_[mr_num]];
 
         if (loopback_ != LoopbackType::LOOPBACK_TYPE_SW) {  // Loopback needs no RX pools
-          if (mr.num_bufs_ < default_num_rx_desc) {
-            DAQIRI_LOG_CRITICAL("Must have at least {} buffers in each RX MR",
-                                  default_num_rx_desc);
+          const uint32_t min_required = (default_num_rx_desc * 3) / 2;  // 1.5x ring size
```
cliffburdick: Why is it 1.5x? Did 1x not work?

cliffburdick: Maybe we should consider just automatically increasing it rather than erroring out. We could increase it with a warning in the logs

dleshchev:
``` 

┌──────────┬───────┬────────┬────────────┬───────────────┬─────────────┐
  │ num_bufs │ ratio │ bursts │ TX packets │ TX bytes / 5s │   Result    │
  ├──────────┼───────┼────────┼────────────┼───────────────┼─────────────┤
  │ 8192     │ 1.00× │ 14     │ 14,336     │ 110 MB        │ 🔒 LOCKED   │
  ├──────────┼───────┼────────┼────────────┼───────────────┼─────────────┤
  │ 9011     │ 1.10× │ 15     │ 15,360     │ 118 MB        │ 🔒 LOCKED   │
  ├──────────┼───────┼────────┼────────────┼───────────────┼─────────────┤
  │ 10240    │ 1.25× │ 7,440  │ 7,627,776  │ 58.7 GB       │ ✅ ~94 Gbps │
  ├──────────┼───────┼────────┼────────────┼───────────────┼─────────────┤
  │ 12288    │ 1.50× │ 7,403  │ 7,591,936  │ 58.4 GB       │ ✅ ~93 Gbps │
  ├──────────┼───────┼────────┼────────────┼───────────────┼─────────────┤
  │ 16384    │ 2.00× │ 7,436  │ 7,628,800  │ 58.6 GB       │ ✅ ~94 Gbps │
  └──────────┴───────┴────────┴────────────┴───────────────┴─────────────┘
```

Did a sweep and found this

dleshchev: updated the code to do an auto-bump to x3 default ring size with a warn for num_bufs < 1.5 default ring size

PR #58, opened 2026-05-06
#56 Fixed failed init leaves hugepage files behind in /dev/hugepages, blocking subsequent runs

### LINKED ISSUE #56: [BUG] Failed init leaves hugepage files behind in /dev/hugepages, blocking subsequent runs
DPDK init failures (e.g. memory regions exceed `HugePages_Free`) leave `<file-prefix>map_*` files in `/dev/hugepages/` that pin every page. `HugePages_Free` stays at 0 and every subsequent run fails the same way until the files are removed by hand. Even clean exit-0 runs leak a few pages.

Hit on `examples/daqiri_bench_raw_tx_rx_spark.yaml`: `num_bufs: 51200` × `buf_size: 8064` overruns the kernel-default 1024 × 2 MiB hugepages. 2048 × 2 MiB works.

1. Boot with kernel-default `nr_hugepages=1024`.
2. `daqiri_bench_raw_gpudirect examples/daqiri_bench_raw_tx_rx_spark.yaml --seconds 5` → fails with `Failed to allocate TX message pool!`.
3. `ls /dev/hugepages/` shows leftover `*map_*` files.
4. Re-run fails identically until `sudo rm -f /dev/hugepages/*map_*`.

**Expected behavior**

- Preflight check on memory-region footprint vs `HugePages_Free` with an actionable error before EAL allocates.
- `rte_eal_cleanup()` on every error path out of `daqiri_init`.
- Surface the manual cleanup recipe from the run instructions (today it's only in a collapsed FAQ entry).

**Environment overview**

- Bare-metal host, bench inside Docker.
- Source build via `scripts/build-container.sh` (`BASE_TARGET=dpdk DAQIRI_MGR="dpdk socket rdma"`).

**Environment details**

- DAQIRI `main` @ 2532bb6 (post PR #45).
- Reproduces with any `kind:` (`huge`, `host_pinned`, `device`) — DPDK uses hugepages for its own mempools/rings regardless.

**Proposed fix (docs-only follow-up to PR #49; in-code work tracked separately)**

- `docs/tutorials/benchmarking_examples.md`, before the first `docker run`: "Configure hugepages first" callout — `grep Huge /proc/meminfo`, size against the YAML's `num_bufs` × `buf_size`, runtime knobs for both 2 MiB and 1 GiB pools, and a link to the persistent grub recipe in `system_configuration.md`.

    ```bash
    echo 2048 | sudo tee /proc/sys/vm/nr_hugepages                                  # 2 MiB pool, 4 GiB
    echo 4    | sudo tee /sys/kernel/mm/hugepages/hugepages-1048576kB/nr_hugepages  # 1 GiB pool, 4 GiB
    ```

- `docs/tutorials/system_configuration.md`, Spark tab: state the shipped Spark YAML needs ~2048 × 2 MiB; default 1024 OOMs at `Failed to allocate TX message pool!`.
- Same file, new "Troubleshooting" sub-section: orphan-hugepage symptom, `<file-prefix>map_*` root cause, cleanup in `sudo` + Docker forms, `pgrep -af daqiri_bench` safety check, leak-on-clean-exit caveat.

    ```bash
    sudo rm -f /dev/hugepages/*map_*
    ```

chloecrozier:
**Fixes #56.** Two things are added:

1. A hugepage check that runs before DPDK starts. If the system does not have enough free hugepages for the YAML config, DAQIRI exits early with a explanation/breakdown and the exact `echo N | sudo tee ...` command to run.
2. A cleanup step that calls `rte_eal_cleanup()` and removes any leftover hugepage files from `/dev/hugepages` and `/mnt/huge` on every init error and on shutdown.

Both helpers live in the `Manager` base class, so anything that uses `libdaqiri.so` — all `daqiri_bench_*`, all `daqiri_example_*`, the Python bindings, and any external program — gets the fix without code changes. Both backends that talk to DPDK (`dpdk` and `rdma`) call the helpers from `initialize()` and `shutdown()`.

## Changes

**Code**

- `src/manager.{h,cpp}`: adds two helpers to the `Manager` base class.
  - `check_hugepage_availability()` reads `/sys/kernel/mm/hugepages` to find how many pages are free, then adds up everything DAQIRI will need: each `kind: HUGE` memory region in the YAML, the dummy queues DAQIRI adds on its own, per-pool overhead, and DPDK's own fixed overhead. If the total is more than what is free, it logs each piece of the sum and prints the exact `echo N | sudo tee ...` command the user should run. Works for both 2 MiB and 1 GiB pages.
  - `cleanup_eal()` calls `rte_eal_cleanup()` and then deletes any leftover `<file-prefix>map_*` files in `/dev/hugepages` and `/mnt/huge`.

- `src/managers/dpdk/daqiri_dpdk_mgr.cpp`:
  - Calls the preflight check at the top of `initialize()`.
  - If a later step in `initialize()` fails before `initialized_ = true` is set, a small guard object runs `cleanup_eal()` automatically as it goes out of scope. This covers every error path without having to add cleanup at each `return`.
  - `shutdown()` now frees the rings **before** `rte_eal_cleanup()` (the previous order was wrong and could touch freed memory).
  - The destructor only frees rings when `eal_initialized_` is true, so it does not touch memory that has already been cleaned up.

- `src/managers/rdma/daqiri_rdma_mgr.cpp`: the same four changes as the DPDK manager — preflight at the top of `initialize()`, scope guard for error paths, ring frees before `rte_eal_cleanup()` in `shutdown()`, and a destructor that checks `eal_initialized_`.

**Docs**

- `docs/tutorials/benchmarking_examples.md`: short "Configure hugepages first" callout pointing at the new check, plus a tighter FAQ pair (preflight path + SIGKILL-orphan cleanup).
- `docs/tutorials/system_configuration.md` (DGX Spark tab): bumped `hugepages=3` → `hugepages=4` in the grub recipe to match the shipped Spark YAML's footprint.

## Tests run

| # | Scenario | Result | Status |
|---|---|---|---|
| 1 | `nr_hugepages=1`, run `daqiri_bench_raw_tx_rx_spark.yaml` | preflight aborts with breakdown + `echo 381 \| sudo tee ...`; `/dev/hugepages` empty | Pass |
| 2 | `nr_hugepages=2048`, run `daqiri_bench_raw_sw_loopback.yaml` | exit 0; `Removed leftover hugepage file ...` lines on shutdown; `HugePages_Free` restored | Pass |
| 3 | Intentioanlly wrong PCI BDF (`0000:ff:ff.0`) to force post-EAL failure | `Failed to get port number`; scope guard runs `rte_eal_cleanup()`; `/dev/hugepages` empty | Pass |
| 4 | Test 1 → Test 2 back-to-back without manual `rm` (the original symptom) | second run starts clean — no manual intervention needed | Pass |
| 5 | `mkdocs build --strict` | clean (0.28s) | Pass |

## How a reviewer can validate

Build inside the container, then walk through the checklist. After each run, check `/dev/hugepages` and `/mnt/huge` are empty (no `map_*` files left behind).

- [ ] **Build** — `cmake -S . -B build -DBUILD_SHARED_LIBS=ON -DDAQIRI_BUILD_PYTHON=OFF -DDAQIRI_MGR="dpdk socket rdma"` then `cmake --build build -j` (run inside the container as the current user). Build succeeds with no errors.
- [ ] **Preflight aborts on too few hugepages** — set `nr_hugepages=1`, then run `daqiri_bench_raw_gpudirect ./examples/daqiri_bench_raw_tx_rx_spark.yaml --seconds 1`. Exit is non-zero, the log shows the per-piece breakdown and an `echo N | sudo tee ...` recipe, `/dev/hugepages` and `/mnt/huge` are empty, and `HugePages_Free` in `/proc/meminfo` is unchanged.
- [ ] **Clean run cleans up after itself** — set `nr_hugepages=2048`, then run `daqiri_bench_raw_gpudirect ./examples/daqiri_bench_raw_sw_loopback.yaml --seconds 5`. Exit is `0`, the log shows `Removed leftover hugepage file ...` lines on shutdown, `/dev/hugepages` and `/mnt/huge` are empty, and `HugePages_Free` returns to `2048`.
- [ ] **Post-EAL failure path** — copy `examples/daqiri_bench_raw_tx_rx_spark.yaml`, replace the PCIe BDFs with bogus values (e.g. `0000:ff:ff.0`), and run the bench against that file. The log shows `Failed to get port number`, the scope guard fires `rte_eal_cleanup()`, and `/dev/hugepages` and `/mnt/huge` are empty.
- [ ] **Back-to-back runs (the original bug)** — run the failing case from the second checkbox, then run the clean case from the third checkbox without any manual `rm`. The second run starts cleanly with no `Cannot allocate memory` or stale-file errors.
- [ ] **Docs build** — `mkdocs build --strict` is clean.


docs/tutorials/benchmarking_examples.md
```diff
@@ -383,6 +393,10 @@ sudo mlnx_perf -i $if_name
 
         Ensure you run as root, using `sudo`.
 
+    ??? failure "Insufficient free hugepages: x MiB free, ~y MiB required for this config."
+
+        DAQIRI's preflight check has caught a hugepage shortfall before EAL was initialized, so no files were left in `/dev/hugepages`. The critical log prints a breakdown (kind:HUGE MRs, dummy queues, per-pool overhead, EAL fixed overhead) and the exact `echo N | sudo tee ...` command to allocate enough pages — copy-paste it. See [Configure hugepages first](#configure-hugepages-first) for the full sizing guide and persistent grub recipe.
```
cliffburdick: We shouldn't mention EAL here since it's an implementation detail.

chloecrozier: I removed the reference and pushed the changes.

cliffburdick (approved):

PR #66, opened 2026-05-08
#65 - Add use-case decision tree to configuration tutorial for picking an example YAML

### LINKED ISSUE #65: [DOC] Add use-case decision tree to configuration tutorial for picking an example YAML
## Report needed documentation

**Report needed documentation**

The `examples/` directory has grown to 19 YAML config files spanning three backends (DPDK raw, RDMA, kernel socket), several data-path specializations (HDS, GPU/CPU reorder with two algorithms, ML quantize, multi-queue), software-vs-hardware loopback, and two on-disk recording paths (PCAP, GDS). A new user landing on `examples/` has no roadmap for which one to start from, and the existing tutorials only show how to run one canonical config (`daqiri_bench_raw_tx_rx.yaml`) without explaining when to pick the others.

The implicit categorization in filenames (`_hds`, `_reorder_seq_*`, `_sw_loopback`, `_spark`) only helps users who already know the vocabulary. There's also no place in the tutorial flow where YAMLs are mapped to the binary that consumes them — 6 of 19 configs run on binaries other than `daqiri_bench_raw_gpudirect`, so a YAML name alone doesn't tell users what to invoke. The mapping table in `examples/README.md` is correct but isn't surfaced from `docs/tutorials/`.

**Describe the documentation you'd like**

A use-case decision tree at the top of `docs/tutorials/configuration-walkthrough.md` that guides users by *what they're trying to do* rather than by feature names they may not recognize. Proposed structure: a top-down triage list of 9 use-case questions phrased in the user's voice — smoke test → GPU packet reordering (flagship) → ML quantize → baseline GPUDirect throughput → HDS → multi-queue → record packet data (PCAP / GDS nested) → RDMA → kernel TCP/UDP sockets. Each leaf names both the YAML and the binary that consumes it, plus the build flags and hardware required.

Backend (DPDK / RDMA / socket) becomes a per-question annotation rather than a top-level structural axis, so the tree starts from user task vocabulary instead of a build choice they may not yet have made. GDS is nested under "I need to record packet data" as an *advanced* sub-option behind PCAP, since it has the strictest prerequisites (`-DDAQIRI_ENABLE_GDS=ON`, NVMe-backed storage, working cuFile / `nvidia_fs` stack).

The existing annotated walkthrough of `daqiri_bench_raw_tx_rx.yaml` stays as the canonical "what does each YAML field mean" reference; the new section sits above it and links into it from the baseline-throughput leaf. A short cross-reference in `docs/tutorials/benchmarking_examples.md` points readers from the benchmark-running tutorial to the decision tree.

**Steps taken to search for needed documentation**

- Read `docs/tutorials/configuration-walkthrough.md` (annotates one YAML; no comparative guidance across the 19 variants).
- Read `docs/tutorials/benchmarking_examples.md` (covers running benchmarks; only Spark-profile YAMLs are explicitly recommended).
- Read `examples/README.md` (has the YAML→binary mapping table but isn't linked from the tutorials).
- Searched `docs/` for any existing decision-tree or selection-guide content — none exists.
- Confirmed against the 19 files in `examples/*.yaml` that the naming convention encodes most of the categorization but doesn't help users who don't know what `_hds` or `_reorder_seq_batch` means.

RamyaGuru:
## Summary

Closes #65. References #4.

The `examples/` directory has 19 YAML configs across three backends and several data-path specializations, but the tutorials only annotated one canonical config and didn't map YAMLs to the binaries that consume them. Adds a use-case decision tree at the top of `docs/tutorials/configuration-walkthrough.md` so a new user can land there and route from "what am I trying to do?" to a specific YAML + binary in a few clicks, without needing to recognize feature names like "HDS" or "reorder" first.

The new section is structured as a top-down list of 5 use-case questions phrased in the user's voice, using collapsible MkDocs Material `???` admonitions:

1. **I want to measure baseline throughput** — backend (DPDK raw / RDMA / sockets) and hardware (generic / Spark) sub-questions; SW loopback nested as a "no physical NIC available" fallback.
2. **I have out-of-order UDP packets that need to be reordered on the GPU** — algorithm (`seq_packets_per_batch` vs. `seq_batch_number`), kernel location (GPU / CPU), direction (TX+RX / RX-only), and an in-kernel type-conversion sub-question (which folds in the former quantize variant).
3. **I need to parse small per-packet metadata on the CPU while keeping payload on the GPU** — HDS.
4. **I need flow-based load balancing across multiple RX queues** — multi-queue.
5. **I need to record packet data to disk** — PCAP (default) and GDS (advanced) nested as 5.1 and 5.2.

Every leaf names the YAML *and* the binary that consumes it (6 of 19 YAMLs run on binaries other than `daqiri_bench_raw_gpudirect`). Backend selection is a per-question annotation rather than a structural axis. The existing annotated walkthrough of `daqiri_bench_raw_tx_rx.yaml` is preserved under a new `## Annotated walkthrough` H2 below the new section.

A one-line cross-reference at the top of `docs/tutorials/benchmarking_examples.md` points readers from the benchmark-running tutorial to the decision tree.

Net change: +93 / −0 lines across 2 files.

## Preview

A preview deploy of this branch is live at https://nvidia.github.io/daqiri/tutorials/configuration-walkthrough/#choosing-an-example-config. The next merge to `main` will overwrite the gh-pages branch and restore the production docs (per the `deploy` job in `.github/workflows/docs.yml`).

## Test plan

- [x] `mkdocs build --strict` passes (CI gate).
- [x] Coverage check: all 19 files in `examples/*.yaml` appear at least once in the rendered tree.
- [x] Binary coverage: all 8 binaries (`daqiri_bench_raw_gpudirect`, `daqiri_bench_raw_hds`, `daqiri_bench_raw_reorder_seq`, `daqiri_bench_raw_reorder_quantize`, `daqiri_example_pcap_writer`, `daqiri_bench_rdma`, `daqiri_bench_socket`, `daqiri_example_gds_write`) appear in leaves.
- [x] Existing annotated walkthrough's highlighted lines and 30 numbered annotations still resolve.
- [x] Cross-link from `benchmarking_examples.md` to `#choosing-an-example-config` resolves.


cliffburdick:
I know this PR is closed, but I wanted to leave a "decision tree" style comment for posterity. DAQIRI has three backends currently that fall under two categories:

1. Raw Ethernet
2. Sockets
 2a. POSIX Sockets
 2b. RoCE

The first part of the decision tree is about which of these to choose:
- Raw Ethernet should be chosen if you have an NVIDIA NIC and you have either a sender or receiver that expects non-connection-oriented raw Ethernet traffic (Ethernet, UDP, etc). 
- POSIX Sockets should be chosen if you do not have an NVIDIA NIC. This is a last resort and will give the worst performance.
- RoCE should be chosen if you control both sides of the connection (server/client) and need reliability guarantees. RoCE requires an implementation of the protocol on top of UDP and must be implemented on both sides.

PR #71, opened 2026-05-11
#70 - Pin cmake-data alongside cmake in Dockerfile base stage

### LINKED ISSUE #70: [BUG] Docker base stage fails: cmake pinned but cmake-data is not
## Summary

`Dockerfile:48-53` pins `cmake` to the latest Kitware-published 3.x version but does not pin `cmake-data` to the same version. As of the Kitware upload on **2026-05-06**, this breaks the `base` stage.

## Repro

```bash
docker build --no-cache --target base --progress=plain \
  --build-arg DAQIRI_OS_BASE_IMAGE=nvcr.io/nvidia/cuda:13.1.0-devel-ubuntu24.04 \
  -t daqiri:repro-base .
```

Fails with:

```
The following packages have unmet dependencies:
 cmake : Depends: cmake-data (= 3.31.12-0kitware2ubuntu24.04.1)
E: Unable to correct problems, you have held broken packages.
```

## Root cause

Kitware's `cmake 3.31.12-0kitware1ubuntu24.04.1` shipped only `Replaces: cmake-data (<< 4.3)` — no hard dependency — so the unpinned install worked.

On **2026-05-06** Kitware re-uploaded the same upstream version as `3.31.12-0kitware2ubuntu24.04.1`, which (re-)introduces `Depends: cmake-data (= 3.31.12-0kitware2ubuntu24.04.1)`. `apt-cache madison cmake | awk '$3 ~ /^3[.]/'` now resolves to the new `kitware2` revision, and without a matching pin on `cmake-data`, apt can't satisfy the dependency (Ubuntu noble's own `cmake-data` is 3.28.x).

## Fix

Pin `cmake-data` to the same version as `cmake` inside the apt install list:

```dockerfile
"cmake=${CMAKE_VERSION}" \
"cmake-data=${CMAKE_VERSION}" \
```

Verified the `base` stage builds cleanly after the change.

dleshchev:
Fixes #70.

## Summary
- Pin `cmake-data` to the same version as `cmake` in the `base` stage so the Kitware-pinned install stays resolvable when Kitware respins a package.
- Add `.github/workflows/docker-base.yml` so a future regression of this kind is caught automatically (cached PR build + daily no-cache drift check on both `ubuntu-24.04` and `ubuntu-24.04-arm`).

## Why this broke now
Previously, `apt-cache madison cmake | awk '$3 ~ /^3[.]/'` resolved to `cmake 3.31.12-0kitware1ubuntu24.04.1`, which only had `Replaces: cmake-data (<< 4.3)` — no hard dependency — so installing just `cmake` worked.

On **2026-05-06** Kitware re-uploaded the same upstream version as `3.31.12-0kitware2ubuntu24.04.1`. That respin (re-)introduces `Depends: cmake-data (= 3.31.12-0kitware2ubuntu24.04.1)`. The madison query now selects the `kitware2` revision, and without a matching pin on `cmake-data` apt cannot resolve the dependency (Ubuntu noble's own `cmake-data` is 3.28.x), so the `base` stage fails with:

```
cmake : Depends: cmake-data (= 3.31.12-0kitware2ubuntu24.04.1)
E: Unable to correct problems, you have held broken packages.
```

## Test plan
- [x] `docker build --no-cache --target base ...` reproduces the failure on `main` (arm64 / Ubuntu 24.04).
- [x] After the one-line change, the same build succeeds and installs `cmake 3.31.12-0kitware2ubuntu24.04.1` + matching `cmake-data`.
- [x] **arm64 (IGX):** full no-cache build through `runtime` target (`DAQIRI_BASE_TARGET=dpdk`, `DAQIRI_MGR="dpdk socket"`) succeeds end-to-end. Stage timing: base 32s · base-deps 22s · dpdk 864s · daqiri-build 77s · runtime 4s → image `daqiri:repro-runtime` (8.41 GB).
- [x] **arm64 (Spark):** full downstream build also succeeds (independently verified on a second arm64 host).
- [x] New `Docker base stage` workflow passes on this PR (cached build on `ubuntu-24.04` and `ubuntu-24.04-arm`).

### DIFF
--- a/.github/workflows/docker-base.yml
+++ b/.github/workflows/docker-base.yml
@@ -1,0 +1,59 @@
+name: Docker base stage
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'Dockerfile'
+      - '.github/workflows/docker-base.yml'
+  pull_request:
+    paths:
+      - 'Dockerfile'
+      - '.github/workflows/docker-base.yml'
+  schedule:
+    # Daily 06:00 UTC drift check against upstream apt repos (Kitware, Ubuntu).
+    # Runs no-cache so cmake / cmake-data and similar pinning regressions are
+    # caught within ~24 hours of an upstream re-upload.
+    - cron: '0 6 * * *'
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: build base (${{ matrix.runner }})
+    strategy:
+      fail-fast: false
+      matrix:
+        runner:
+          - ubuntu-24.04
+          - ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: docker/setup-buildx-action@v3
+
+      - name: Build base stage (cached, PR/push)
+        if: github.event_name != 'schedule'
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          target: base
+          push: false
+          build-args: |
+            DAQIRI_OS_BASE_IMAGE=nvcr.io/nvidia/cuda:13.1.0-devel-ubuntu24.04
+          cache-from: type=gha,scope=base-${{ matrix.runner }}
+          cache-to: type=gha,mode=max,scope=base-${{ matrix.runner }}
+
+      - name: Build base stage (no-cache, scheduled drift check)
+        if: github.event_name == 'schedule'
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          target: base
+          push: false
+          no-cache: true
+          build-args: |
+            DAQIRI_OS_BASE_IMAGE=nvcr.io/nvidia/cuda:13.1.0-devel-ubuntu24.04
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && apt-get install --no-install-recommends -y \
         kitware-archive-keyring \
         "cmake=${CMAKE_VERSION}" \
+        "cmake-data=${CMAKE_VERSION}" \
     && rm -rf /var/lib/apt/lists/*
 
 # ==============================================================


RamyaGuru (commented):
Should add a simple CI script to test the build? It sounds like "docker build --no-cache --target base-deps" might be lightweight enough to run through the free Github Actions cloud resources?

src/managers/dpdk/daqiri_dpdk_mgr.cpp
```diff
@@ -1581,8 +1581,19 @@ void DpdkMgr::adjust_memory_regions() {
   const uint32_t deadlock_threshold = (ring_size * 3) / 2;  // 1.5x ring size
   const uint32_t bumped_num_bufs    = ring_size * 3;        // 3x ring size
 
+  std::unordered_set<std::string> queue_backed_mrs;
+  for (const auto& intf : cfg_.ifs_) {
+    for (const auto& q : intf.rx_.queues_) {
+      for (const auto& n : q.common_.mrs_) { queue_backed_mrs.insert(n); }
```
cliffburdick: Is it intended to share a set for both TX and RX MRs?

dleshchev:
Rebased onto `main` so this PR is now a standalone one-line fix plus the new CI workflow. The two commits from #68 (`#44`, `#56`) are no longer in this PR's diff.

@cliffburdick — your review comment on `src/managers/dpdk/daqiri_dpdk_mgr.cpp:1587` was on the change introduced by #68 (the `queue_backed_mrs` set populated from both `rx_.queues_` and `tx_.queues_`). I'll re-post your question as an inline comment on #68 so the discussion continues there. Sorry for the routing churn — earlier I'd branched this fix off #68 because it was the only branch handy that built clean.

dleshchev:
@RamyaGuru — added `.github/workflows/docker-base.yml` in [`b3e8133`](https://github.com/NVIDIA/daqiri/commit/b3e8133). Two triggers:

- **PR / push to main**: cached `docker build --target base` on `ubuntu-24.04` and `ubuntu-24.04-arm` (free runners now that the repo is public). Per-PR cost is ~30s after the first run thanks to the GHA layer cache.
- **Nightly (06:00 UTC)**: same target with `no-cache: true`. This is the actual drift guard — it forces apt to re-resolve, so a future Kitware respin (or any equivalent break in the Ubuntu base) would page within 24 hours instead of waiting for someone to notice locally.

Scoped to `--target base` only — that's the stage that pins apt versions and exercises Kitware's repo. The DPDK source build and DAQIRI compile happen in later stages and aren't necessary to catch this class of drift; we can layer those in later if it ever feels worth the ~20 min runner cost.

RamyaGuru (approved):

PR #72, opened 2026-05-11
#15 - Add C++ bench infrastructure for DGX Spark performance report

RamyaGuru:
## Summary

  Lays the C++ groundwork for the DGX Spark performance report (#15). Code and doc-skeleton only; per-cell numbers fill in via a follow-on commit on this same PR before it's marked ready-for-review.

  - Adds `TokenBucketPacer` (`--target-gbps` software pacer) and `seconds=` field on bench output across DPDK / RoCE / Socket benches. Zero `src/` changes.
  - Adds `examples/bench_capture_environment.sh` (slow-moving system state snapshot) and `examples/run_spark_bench.sh` (sweep wrapper that runs the bench under `mpstat` + `nvidia-smi dmon` and emits one CSV row per
   cell).
  - Adds `docs/performance-dgx-spark.md` skeleton with full final structure: native-shape and matched 8 KB headline tables (so cross-backend Gbps comparisons are honest), per-backend sweep dimensions (DPDK packets
  / RoCE messages / TCP stream), and explicit "HDS deferred on GB10" note.
  - Wires the doc into `mkdocs.yml`, `docs/index.html` News card, `README.md`, and `AGENTS.md` drift-hotspot list.

  Drop accounting routes through the wrapper (no manager changes): DPDK via `DAQIRI_LOG_INFO` parsing, RDMA via `"CQ error"` lines, Socket UDP via `/proc/net/udp` diff, Socket TCP via `nstat`.

  ## Test plan

  - [x] DPDK smoke (`daqiri_bench_raw_gpudirect`): TX/RX `complete:` lines have the new `seconds=` format; ~84 Gbps cross-chip on Spark, 0 drops.
  - [x] Pacer accuracy: `--target-gbps 1` holds 1.055 Gbps (within the +/-5% software pacer accuracy documented in the report).
  - [x] Socket UDP smoke (`daqiri_bench_socket`): new `sent_packets`/`recv_packets`/`sent_bytes`/`recv_bytes`/`seconds=` format; iteration cap behavior preserved.
  - [x] `bench_capture_environment.sh` captures kernel cmdline, CPU, NUMA, hugepages, mlx5 PCIe, OFED, NVIDIA GPU state, isolcpus, governor, IRQ.
  - [x] `run_spark_bench.sh dpdk smoke` produces a valid CSV row with `gbps~=96`, `drops=0`, drops_kind tagged correctly.
[...]

PR #85, opened 2026-05-15
#75 - Fix Socket UDP --mode both; error spam impacts startup

### LINKED ISSUE #75: [BUG] Socket UDP --mode both deadlocks on peer learning; only ~1000 packets / 30 s trickle through
## Symptom

  Running `daqiri_bench_socket` against `examples/daqiri_bench_socket_udp_tx_rx.yaml` with `--mode both --seconds 30` produces ~1000 packets and ~390 kbps effective throughput. Stderr spams the following for the
  duration of the run:

  ```
  [ERROR] /workspace/daqiri/src/managers/socket/daqiri_socket_mgr.cpp:735: UDP server has no learned peer yet; cannot transmit
  ```

  Reproduced on DGX Spark (GB10) during PR #15 / commit `5e57a5b` data-fill on 2026-05-12.

  ## Repro

  ```bash
  ./build/examples/daqiri_bench_socket \
      examples/daqiri_bench_socket_udp_tx_rx.yaml \
      --seconds 30 --mode both
  ```

  ## Root cause

  Both server and client workers attempt to send before either has received an inbound packet. UDP is connectionless, so the server has no peer address to send to until the client's first packet arrives. With both
  ends spinning in send-then-receive in the same process, only a trickle of packets gets through and the server's `[ERROR]` log spam masks the real throughput.

  ## Fix sketch (one or both)

  - In the bench (`examples/socket_bench.cpp`), defer the server's send loop until the first inbound packet has populated the peer address.
  - In the manager (`src/managers/socket/daqiri_socket_mgr.cpp:735`), downgrade the "no learned peer" log to DEBUG once-per-peer. It's currently CRITICAL/ERROR once per attempt — ~30 lines per second of attempted
  send.

  ## Scope

  `src/managers/socket/daqiri_socket_mgr.cpp` + `examples/socket_bench.cpp`. Unblocks the Socket UDP rows of `docs/performance-dgx-spark.md`.

  ## Context

  This issue was uncovered during PR #15 verification on DGX Spark. The Socket UDP rows in the performance report are footnoted as `_deferred_[^2]` pending this fix.

RamyaGuru:
Fixes #75.

_**NOTE TO REVIEWER:** the original Git issue implies that this caused a throughput cliff in benchmarks for socket UDP. In reality, this was a red herring and it was an issue with the benchmark test. So, this is more of a cosmetic change to fix some error-spamming that also leads some sluggishness on startup (see below)._

  The server thread, when invoked in `--mode both`, tries to send before any inbound packet has arrived. With no learned peer, each `send_udp_burst` returns false after logging `UDP server has no learned peer yet; cannot transmit` at
  ERROR. The spam stops once the first inbound packet teaches the server its peer (typically within ~30 attempts), but each ERROR write through spdlog→stderr is a synchronous syscall that holds `state_mutex_` and contends with the
  client's first packets.

  Two changes:
  1. **`examples/socket_bench.cpp`** — gate the server worker's TX path on having received at least one packet (only when `receive` is enabled), so the client transmits eagerly and the server learns the peer before it tries to send.
  2. **`src/managers/socket/daqiri_socket_mgr.{h,cpp}`** — throttle the "no learned peer" log to fire once per missing-peer episode and demote it from ERROR to DEBUG. A new `udp_peer_missing_logged` latch on `EndpointState` is cleared
  when the RX loop learns/relearns a peer.

  In pathological cases where the server never receives an inbound packet (one-way flow, mis-bound RX queue), the original code would spam ERROR forever; the latch makes those scenarios survivable too.

  ## Test plan

  UDP loopback (`127.0.0.1`), `--seconds 30 --mode both`, message_size=1024, iterations bumped to 100M so the wall-clock cap dominates. `bench-results/` runs on a DGX Spark in container build.

  - [x] **Revert + reproduce** — checked out `main`'s versions of the three modified files, rebuilt, ran the bench. Confirmed 33 `no learned peer` ERROR lines on the broken side.
  - [x] **Re-apply + verify symptom gone** — restored the fix, rebuilt, re-ran. ERROR count: **33 → 0**.
  - [x] **N=3 throughput, total tx_pkts / 30s, both endpoints:**

    |          | Trial 1 | Trial 2 | Trial 3 | Min | Max | **Mean** |
    |----------|--------:|--------:|--------:|----:|----:|---------:|
    | Broken   | 17.67M  | 16.57M  | 17.44M  | 16.57M | 17.67M | **17.23M** |
    | Fixed    | 18.89M  | 18.47M  | 18.50M  | 18.47M | 18.89M | **18.62M** |

    Mean delta: **+1.39M pkts / 30s (+8.1%)**. Ranges do not overlap — slowest fixed trial still beats fastest broken trial.

  - [x] Code audit of `udp_peer_valid` in `main`'s `daqiri_socket_mgr.cpp`: set true once in the RX loop after the first `recvmmsg`, never cleared — so the broken-side spam is bounded by time-to-first-inbound-packet, not run length.
  Matches the observed ~33 ERROR count being identical in iter=1000 and iter=100M runs.

### DIFF
--- a/examples/socket_bench.cpp
+++ b/examples/socket_bench.cpp
@@ -96,7 +96,12 @@ void socket_worker(const SocketBenchConfig& cfg, std::atomic<bool>& stop, Socket
     const bool recv_done = !cfg.receive || stats.received_packets >= static_cast<uint64_t>(cfg.iterations);
     if (send_done && recv_done) { break; }
 
-    if (cfg.send && !send_done) {
+    // UDP servers must learn the peer address from an inbound packet before
+    // they can transmit, so withhold the server's first send until at least
+    // one RX packet has arrived. The client always transmits eagerly.
+    const bool waiting_for_peer = cfg.server && cfg.receive && stats.received_packets == 0;
+
+    if (cfg.send && !send_done && !waiting_for_peer) {
       auto* msg = daqiri::create_tx_burst_params();
       daqiri::set_header(msg, port, queue, 1, 1);
 
--- a/src/managers/socket/daqiri_socket_mgr.cpp
+++ b/src/managers/socket/daqiri_socket_mgr.cpp
@@ -739,7 +739,10 @@ bool SocketMgr::send_udp_burst(EndpointState& ep, BurstParams* burst, size_t* se
   if (ep.socket_cfg.mode_ == SocketMode::SERVER) {
     std::lock_guard<std::mutex> lock(state_mutex_);
     if (!ep.udp_peer_valid) {
-      DAQIRI_LOG_ERROR("UDP server has no learned peer yet; cannot transmit");
+      if (!ep.udp_peer_missing_logged) {
+        DAQIRI_LOG_DEBUG("UDP server has no learned peer yet; cannot transmit");
+        ep.udp_peer_missing_logged = true;
+      }
       return false;
     }
     peer = ep.udp_peer_addr;
@@ -1285,6 +1288,7 @@ void SocketMgr::udp_rx_loop(int if_index) {
       std::lock_guard<std::mutex> lock(state_mutex_);
       ep->udp_peer_addr = peers[static_cast<size_t>(received - 1)];
       ep->udp_peer_valid = true;
+      ep->udp_peer_missing_logged = false;
     }
 
     for (int i = 0; i < received; ++i) {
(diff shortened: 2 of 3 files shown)


RamyaGuru:
Closing this — issue #75 turned out to be a red herring. There's no actual deadlock in Socket UDP --mode both; the benchmark just stops after 1000 packets (the default iterations cap), which made the brief peer-learning warmup look like a hang.

  What the branch actually fixed was cosmetic: the server side eagerly calls send_udp_burst before the first RX packet arrives, so udp_peer_valid is false for a handful of iterations and each failed send logs an ERROR. Once one client
  packet lands, the peer is learned and sending proceeds normally.

  The two-line log cleanup (downgrade ERROR → DEBUG and rate-limit to once) is worth keeping but doesn't justify its own PR — I'll fold it into #72 if it's still relevant there. Closing this PR and issue #75.

Issue #15, opened 2026-04-16
[FEA] Benchmark DAQIRI on DGX Spark (C++)


awthomp:
**Describe the solution you'd like**
Backend - DPDK, HDS
- Loopback
- Default yaml config file + FFT
- Default yaml config file + GEMM

Backend - RoCE
- Loopback
- Default yaml config file + FFT
- Default yaml config file + GEMM

Show:
- Total data rate
- Total packets sent
- Total packets received
- Total dropped

PR #96, opened 2026-05-26
#15 - Add reusable C++ bench infrastructure for per-platform performance reports

RamyaGuru:
## Summary

First of a 3-PR stack replacing **#72** (DGX Spark performance report). This PR contains only the platform-agnostic bench plumbing — no Spark-specific tooling, no perf doc.

- **`TokenBucketPacer`** in `examples/raw_bench_common.{cpp,h}`, plus a `--target-gbps G` CLI flag that wires it into the raw / RDMA / socket benches.
- **`--seconds N`** CLI flag for time-bounded runs (replaces the previous YAML `iterations` requirement for smoke runs).
- **Atomic stringstream prints** so the multi-thread bench threads don't garble their `RX complete:` / `TX complete:` lines.
- **Drop-count propagation** — DPDK manager xstats summary, RDMA `CQ error` lines, and socket bench `sent_packets`/`sent_bytes` keys are all emitted in a grep-friendly format that the upstream sweep wrapper consumes.
- **`examples/bench_capture_environment.sh`** — one-shot capture of slow-moving host state (kernel, NIC firmware, IOMMU, hugepages, GPU info) into a results directory.

Used by the upcoming DGX Spark, IGX, and x86 platform performance reports (issues #15, #16, and follow-on platform PRs).

## Stack

This PR → **#β (Spark sweep tooling)** → **#γ (Spark perf doc)** — opening next once their test plans pass.

Replaces #72 (which had all three concerns bundled into one ~1,545-line draft).

## Test plan

- [x] Clean build: `cmake -S . -B build -DBUILD_SHARED_LIBS=ON -DDAQIRI_BUILD_PYTHON=OFF -DDAQIRI_MGR="dpdk socket rdma" && cmake --build build -j` — no warnings.
- [x] `--seconds` honored on DPDK QSFP loopback (`daqiri_bench_raw_gpudirect` against `daqiri_bench_raw_tx_rx_spark.yaml`) — TX/RX complete lines print cleanly, RX packets > 0, exits in ~5 s.
- [x] `TokenBucketPacer` rate-limits: `--target-gbps 10` measured within ±20% of the target on the same cable.
- [ ] Socket bench output format (`Server complete: sent_packets=... sent_bytes=...`) — deferred to PR β's regression test, which exercises the same field-name path through the sweep wrapper.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -98,14 +98,19 @@ add_daqiri_raw_bench(daqiri_bench_raw_reorder_quantize raw_reorder_quantize_benc
 add_daqiri_raw_bench(daqiri_example_gds_write gds_write_example.cpp)
 add_daqiri_raw_bench(daqiri_example_pcap_writer pcap_writer_example.cpp)
 
-add_executable(daqiri_bench_rdma rdma_bench.cpp)
+add_executable(daqiri_bench_rdma rdma_bench.cpp raw_bench_common.cpp)
 link_daqiri_bench(daqiri_bench_rdma)
+target_link_libraries(daqiri_bench_rdma PRIVATE CUDA::cudart)
 set_target_properties(daqiri_bench_rdma PROPERTIES
   BUILD_RPATH "$ORIGIN/../src;$ORIGIN/../src/third_party/yaml-cpp"
 )
 
-add_executable(daqiri_bench_socket socket_bench.cpp)
+add_executable(daqiri_bench_socket socket_bench.cpp raw_bench_common.cpp)
 link_daqiri_bench(daqiri_bench_socket)
+target_link_libraries(daqiri_bench_socket PRIVATE CUDA::cudart)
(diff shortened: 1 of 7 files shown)


RamyaGuru:
Force-pushed to add a missing `BUILD_RPATH` on `daqiri_bench_socket` in `examples/CMakeLists.txt`. The original commit linked `raw_bench_common.cpp` + `CUDA::cudart` into the socket bench (matching what was done for `daqiri_bench_rdma`) but missed the matching `BUILD_RPATH` property. Without it, the binary runtime-links against `/opt/daqiri/lib/libdaqiri.so` — the installed snapshot, which can be older than the source tree (it was, in this case: missing Cliff's `set_connection_id` accessors from #88). Found while running PR #97's regression test, which crashed with an undefined-symbol error until I added the RPATH. `daqiri_bench_rdma` already had the same property, so this is just bringing socket in line.

PR #97, opened 2026-05-26
#15 - Add DGX Spark sweep tooling and RDMA loopback prereqs

RamyaGuru:
## Summary

Second of a 3-PR stack replacing **#72**. Layers DGX Spark-specific sweep tooling and the RDMA-bench depth bump on top of #96 (reusable bench infra).

- **`examples/run_spark_bench.sh`** — Spark-tuned sweep driver with per-backend payload/batch matrices, CPU pins, drop-source dispatch (DPDK imissed/ierrors/nombuf, RDMA `CQ error`, socket `/proc/net/udp` drops + `nstat` retrans), and one CSV row per cell into `bench-results/`.
- **`scripts/spark_data_fill.sh`** — one-shot driver that runs the full bench matrix across DPDK / socket-UDP / socket-TCP, with hugepage pre-flight and orphan-hugepage cleanup between runs.
- **`scripts/setup_spark_rdma_loopback.sh`** — idempotent host prereq pinning static ARP entries and source-based policy routing for the Spark single-NIC cross-cable RoCE loopback. The script hardcodes the Spark CX-7 netdev names, MACs, and 1.1.1.1 / 2.2.2.2 IPs from `daqiri_bench_rdma_tx_rx_spark.yaml` — the technique generalizes to any single-NIC RoCE loopback, the script does not (the `spark_` prefix is intentional).
- **`examples/rdma_bench.cpp`** — raise `kMaxOutstanding` 5→20 to match `num_bufs` in the YAML configs. Lifts small-payload pps 8–22× on Spark (4 KB: 4→39 msg/s, 8 KB: 4→88, 64 KB: 32→255) without affecting 8 MB / 1 MB cells already saturated at depth 5. The new comment documents why the constant cannot exceed `num_bufs` (post_req / free_tx_burst ordering in the same loop iteration would deadlock instead of throttling). A follow-up tracks the deeper architectural fix.
- **`.gitignore`** — `pcie_schematic.png` (generated by `tune_system.py`).

Also includes two parse fixes Greptile flagged on the draft of `run_spark_bench.sh`:

- `/proc/net/udp` column 13 (drops) is decimal (`%lu` in `net/ipv4/udp.c`), not hex. Dropped the `strtonum("0x" ...)` treatment that was silently multiplying drop counts whenever the column value contained any digit > 9.
- The socket bench emits `sent_packets` / `sent_bytes`, not RDMA's `send_completions` / `send_bytes`, so the RDMA-keyed fallback was always returning empty for socket backends and producing zero-filled CSV rows. The fallback now dispatches on `$BACKEND`.

## Stack

**Stacked on #96 — must land after.**

PR γ (Spark perf doc) opens next once its checks pass.

## Test plan

- [x] Re-build clean on top of β: `cmake --build build -j` — no warnings.
- [x] `setup_spark_rdma_loopback.sh` idempotent: two consecutive `sudo ./scripts/setup_spark_rdma_loopback.sh` runs both exit 0; `ip rule list` shows 4 rules and `arp -n` shows 4 PERM entries for 1.1.1.1 / 2.2.2.2.
- [x] Greptile field-name fix regression test: `./examples/run_spark_bench.sh socket-udp smoke` now produces a CSV row with `packets=1000, bytes=1472000` (pre-fix the same run produced all zeros). Throughput is low because socket-UDP at MTU is bench-bottlenecked on Spark — that's a separate characterization concern for the perf-doc data-fill follow-up, not a wrapper bug.
- [x] kMaxOutstanding bump + RDMA smoke: `./examples/run_spark_bench.sh rdma smoke` produces `gbps=83.764` on the 8 MB native cell (matches the perf-doc headline) with 0 CQ errors.
- [x] `/proc/net/udp` decimal-parse sanity: no order-of-magnitude inflation between the host-side `awk` sum and the CSV drops field.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,9 @@
 build*/
 site/
+bench-results/
+
+# tune_system.py default output
+pcie_schematic.png
 
 # macOS
 .DS_Store
--- a/examples/rdma_bench.cpp
+++ b/examples/rdma_bench.cpp
@@ -67,12 +67,10 @@ RdmaBenchConfig parse_rdma_cfg(const YAML::Node& node) {
 
 void rdma_worker(const RdmaBenchConfig& cfg, daqiri::bench::TokenBucketPacer& pacer,
                  std::atomic<bool>& stop, RdmaWorkerStats& stats) {
-  // Matches the per-MR num_bufs in the YAML configs. Higher values deadlock
-  // the bench: post_req blocks in get_tx_packet_burst when the pool is empty,
-  // but free_tx_burst (which refills it) only runs later in the same loop
-  // iteration via get_rx_burst. Until the loop is refactored to interleave
-  // drain with post, this constant must stay <= num_bufs.
-  static constexpr int kMaxOutstanding = 20;
+  // Application credit limit. Buffer exhaustion is handled as backpressure so
+  // this can be larger than a memory region's num_bufs: post_req returns to the
+  // outer loop, which drains completions and releases buffers before retrying.
+  static constexpr int kMaxOutstanding = 64;
   int outstanding_send = 0;
   int outstanding_recv = 0;
   uint64_t send_wr_id = 0x1234;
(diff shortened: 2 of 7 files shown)


dleshchev:
Clarifying stack repair state as of 2026-06-03: this PR shows MERGED because it was merged into the stacked base branch `15-bench-infra`, not because merge commit `73cabc97a29fa3349b6611d8fb3eacf3df5f8373` is the mainline landing path. The main-bound content was rebuilt on top of current `main` in follow-up PR #116, which has now landed on `main`.

Cleanup note: once PR #118 lands, the stale stack branches under `15-bench-*` / `cburdick/*` can be pruned. Keep `backup/15-bench-infra-stacked-20260603` until #118 is safely landed.

PR #107, opened 2026-05-29
#106 - Fix CUDA 12.6 raw benchmark build

### LINKED ISSUE #106: [BUG] Raw benchmarks fail to build with CUDA 12.6
**Describe the bug**
DAQIRI raw benchmark examples do not build with CUDA Toolkit 12.6 because
`examples/raw_bench_common.cpp` depends on CUDA Runtime APIs that are not present
in the CUDA 12.6 headers:

- `cudaMemcpyAttributes`
- `cudaMemcpySrcAccessOrderStream`
- `cudaMemcpyFlagDefault`
- `cudaMemcpyBatchAsync`

This blocks the bare-metal build path for the benchmark examples on systems that
use CUDA 12.6.

On a CUDA 12.6 host:

```bash
cmake -S . -B build \
  -DCMAKE_BUILD_TYPE=Release \
  -DBUILD_SHARED_LIBS=ON \
  -DDAQIRI_MGR="dpdk socket rdma" \
  -DDAQIRI_BUILD_EXAMPLES=ON \
  -DDAQIRI_BUILD_PYTHON=OFF
cmake --build build -j"$(nproc)"
```

The build fails while compiling the raw benchmark shared helper because the CUDA
12.6 headers do not declare the batch memcpy API symbols listed above.

**Expected behavior**
DAQIRI raw benchmark examples should build with CUDA Toolkit 12.6 and newer
without requiring users to edit source files or disable the affected benchmark
targets.

 - Environment location: Bare-metal
 - Method of DAQIRI install: source

**Environment details**
 - OS and version: IGX bare-metal development environment
 - Python version: not applicable
 - DAQIRI version: current `main`
 - Hardware configuration: NVIDIA IGX + RTX 6000 Ada
 - CUDA Toolkit: 12.6.68

**Additional context**
The DAQIRI container currently uses a newer CUDA base image where these symbols
are available, so this shows up specifically in the bare-metal CUDA 12.6 path.
The examples do not need `cudaMemcpyBatchAsync`; a custom stream-ordered CUDA
copy helper can provide the needed behavior while keeping the benchmark path
compatible with CUDA 12.6.

dleshchev:
## Summary

Closes #106.

This PR removes the raw benchmark examples' dependency on
`cudaMemcpyBatchAsync`, which is not available in CUDA Toolkit 12.6. The public
example helper signature stays unchanged:

```cpp
cudaError_t memcpy_batch_async(const std::vector<void*>& dsts,
                               const std::vector<const void*>& srcs,
                               const std::vector<size_t>& sizes,
                               cudaStream_t stream)
```

The replacement lives in `examples/raw_bench_common_cuda.cu` and uses:

- a `cudaMemcpy2DAsync` fast path for regular strided batches,
- a small custom CUDA copy kernel for irregular batches with kernel-accessible
  pointers,
- a stream-ordered `cudaMemcpyAsync` loop fallback for cases the kernel should
  not handle.

It handles empty input, mismatched vector sizes, zero-sized copies, and null
nonzero copies. The implementation keeps async stream semantics and avoids CUDA
Runtime APIs newer than CUDA 12.6.

## CMake / CUDA Architecture Notes

This PR also adjusts CUDA architecture defaults. This is not the core
`cudaMemcpyBatchAsync` replacement, but it is needed for the same bare-metal
CUDA 12.6 build path: the previous hardcoded `80;90;121` list makes CUDA 12.6
try to compile `sm_121`, which that toolkit does not support.

New behavior:

- default CUDA architectures are `80;90`,
- `121` is appended only when configuring with CUDA compiler 13.0 or newer,
- user-provided `CMAKE_CUDA_ARCHITECTURES` still takes precedence,
- examples set CUDA architectures explicitly because they now compile a `.cu`
  file and can also be built as their own CMake project.

`docs/getting-started.md` is updated to match this behavior.

## Validation

Formatting:

```bash
git clang-format --style file
git diff --check
```

Bare-metal CUDA 12.6.68 configure/build:

```bash
cmake -S . -B /tmp/daqiri-cuda126-review-fresh \
  -DCMAKE_BUILD_TYPE=Release \
  -DBUILD_SHARED_LIBS=ON \
  -DDAQIRI_MGR="dpdk socket rdma" \
  -DDAQIRI_BUILD_EXAMPLES=ON \
  -DDAQIRI_BUILD_PYTHON=OFF
cmake --build /tmp/daqiri-cuda126-review-fresh \
  --target daqiri_bench_raw_reorder_quantize -j"$(nproc)"
```

Verified CUDA 12.6 emitted only `compute_80/sm_80` and `compute_90/sm_90`, with
no `sm_121`.

Bare-metal CUDA 12.6.68 runtime validation:

```bash
sudo /tmp/daqiri-cuda126-review-fresh/examples/daqiri_bench_raw_reorder_quantize \
  examples/daqiri_bench_raw_tx_rx_reorder_quantize_seq_batch_filled.yaml \
  --seconds 10
```

Result:

```text
Reorder quantize verify: input=int4 output=fp32 verified_batches=3 verified_chunks=18 failures=0
Missed packets: 0
Errored packets: 0
RX out of buffers: 0
```

Container CUDA 13.1 build:

```bash
docker run --rm --privileged --gpus all --network host --user 1000:1001 \
  -v /dev/hugepages:/dev/hugepages \
  -v /tmp:/hosttmp \
  -v /home/dleshchev/repos/daqiri:/workspace/daqiri \
  -w /workspace/daqiri daqiri:dev \
  bash -lc 'cmake -S . -B /hosttmp/daqiri-container-build-review-fresh2 \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_SHARED_LIBS=ON \
    -DDAQIRI_MGR="dpdk socket rdma" \
    -DDAQIRI_BUILD_EXAMPLES=ON \
    -DDAQIRI_BUILD_PYTHON=OFF &&
    cmake --build /hosttmp/daqiri-container-build-review-fresh2 \
      --target daqiri_bench_raw_reorder_quantize -j$(nproc)'
```

Also built the other raw targets that now compile the new `.cu` helper:

- `daqiri_bench_raw_hds`
- `daqiri_bench_raw_gpudirect`
- `daqiri_bench_raw_reorder_seq`
- `daqiri_example_gds_write`
- `daqiri_example_pcap_writer`

Verified CUDA 13.1 emitted `80;90;121`.

Container runtime validation:

```text
Reorder quantize verify: input=int4 output=fp32 verified_batches=3 verified_chunks=18 failures=0
```

## Performance

Affected no-validation throughput path:

```bash
/hosttmp/daqiri-container-build-review-fresh2/examples/daqiri_bench_raw_reorder_quantize \
  /hosttmp/daqiri_reorder_quantize_noverify.yaml \
  --seconds 10
```

Latest patched run:

```text
rx_phy_packets: 64,029,696
rx_phy_bytes: 20,745,621,504
approx rx phy throughput: 16.6 Gbps over 10s
Missed packets: 0
Errored packets: 0
RX out of buffers: 0
verified_batches=0 verified_chunks=0 failures=0
```

Earlier comparison on the same host/config:

- old `cudaMemcpyBatchAsync` path: ~13.1M packets, ~3.4 Gbps phy
- patched path: ~64M packets, ~16.6-16.8 Gbps phy

Large-packet reorder line-rate sanity checks also reached about 94 Gbps phy in
the patched build, matching the expected raw/reorder baseline for that config.


dleshchev:
Addressed Greptile's AGENTS.md note in f9b33f0: the CUDA architecture guidance now matches the conditional defaults (80;90, plus 121 with CUDA >= 13.0).

examples/raw_bench_common_cuda.cu
```diff
@@ -0,0 +1,303 @@
...
+cudaError_t memcpy_async_loop(const std::vector<void *> &dsts,
+                              const std::vector<const void *> &srcs,
+                              const std::vector<size_t> &sizes,
+                              cudaStream_t stream) {
+  for (size_t i = 0; i < dsts.size(); ++i) {
+    if (sizes[i] == 0) {
+      continue;
+    }
+    const cudaError_t status =
+        cudaMemcpyAsync(dsts[i], srcs[i], sizes[i], cudaMemcpyDefault, stream);
```
cliffburdick: I'm not sure what path this was used on before, but a looped memcpy will be very slow if the loop count is large. This could instead be a copy kernel that mimics the batch async copy behavior.

cliffburdick (changes requested):

examples/raw_bench_common_cuda.cu
dleshchev: Addressed in fea9fa7. The kernel path was already the intended path for the benchmark mapped pinned-host staging buffer ([cudaHostAllocMapped](https://github.com/NVIDIA/daqiri/blob/fea9fa79f9880ac7399b44c999c7d14875dca67e/examples/raw_bench_common.cpp#L215-L217)) -> GPU packet-buffer copies ([copy source/destination setup and memcpy_batch_async call](https://github.com/NVIDIA/daqiri/blob/fea9fa79f9880ac7399b44c999c7d14875dca67e/examples/raw_reorder_quantize_bench.cpp#L715-L751)), so I removed the looped cudaMemcpyAsync fallback rather than trying to optimize it.

If a pointer cannot be made kernel-accessible, memcpy_batch_async now returns the CUDA error directly, and device op-buffer allocation failure also returns the allocation status instead of silently degrading to the slow loop.

Verified with the container build and both software-loopback and physical NIC reorder-quantize smoke tests:
Reorder quantize verify: input=int4 output=fp32 verified_batches=3 verified_chunks=18 failures=0

cliffburdick (approved):

PR #110, opened 2026-06-01
#15 - Add Spark socket wire benchmark notes

cliffburdick:
This PR documents how to run Linux networking loopback tests (RDMA/Sockets) on the DGX Spark platform. Without configuring the system like this the benchmarks are not realistic since the traffic is looped through Linux rather than leaving on the wire.

### DIFF
--- a/SPARK_SOCKET_WIRE_README.md
+++ b/SPARK_SOCKET_WIRE_README.md
@@ -1,0 +1,414 @@
+# Spark Socket Wire Benchmark Notes
+
+This note documents the setup needed to make DAQIRI socket benchmark traffic
+leave the NIC and return through a physical receive path on a NVIDIA DGX Spark
+style machine.
+
+Linux network namespaces are not enough by themselves. On Spark, some pairs of
+mlx5 netdevs can exchange traffic through an internal NIC or e-switch path. In
+that case DAQIRI and vport counters move, but physical counters such as
+`rx_packets_phy` do not. Treat benchmark results as "on wire" only after the
+directional PHY counters move.
+
+## Summary
+
+The pair below was verified to send packets out through the physical path:
+
+```text
+client TX: enp1s0f0np0
+server RX: enp1s0f1np1
+```
+
+The pair below did not hit the physical receive path on the tested Spark
+machine:
+
+```text
+client TX: enp1s0f0np0
+server RX: enP2p1s0f0np0
+```
+
+For one-way client to server traffic, the counters that must increase are:
+
+```text
+client interface: tx_packets_phy, tx_bytes_phy
+server interface: rx_packets_phy, rx_bytes_phy
+```
+
+The client `tx_packets_phy` delta should match the server `rx_packets_phy`
+delta, apart from a few control packets. The byte counters should also match.
+
+## Required DAQIRI Socket Changes
+
+Use a DAQIRI build that includes the socket backend and the socket changes from
+this branch:
+
+- UDP payloads above `65507` bytes are rejected, matching iperf's UDP payload
+  limit.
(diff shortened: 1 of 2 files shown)


SPARK_SOCKET_WIRE_README.md
```diff
@@ -0,0 +1,414 @@
+# Spark Socket Wire Benchmark Notes
```
dleshchev, line 1: should this file live in tutorials?

SPARK_SOCKET_WIRE_README.md
```diff
@@ -0,0 +1,414 @@
...
+Set variables for the verified wire-facing pair:
+
+```bash
+CLIENT_NS=dq_wire_client
+SERVER_NS=dq_wire_server
+
+CLIENT_IF=enp1s0f0np0
+SERVER_IF=enp1s0f1np1
+
+CLIENT_IP=10.250.0.1
```
dleshchev, line 134: may be needs a note that this is device-specific?

SPARK_SOCKET_WIRE_README.md
```diff
@@ -0,0 +1,414 @@
...
+count. That proves traffic is entering the server NIC from the physical path.
+
+If only `tx_vport_unicast_*` and `rx_vport_unicast_*` move, but `rx_packets_phy`
+does not, the traffic is not on the wire. Pick another port pair.
+
+## DAQIRI Socket Configs
+
+The socket benchmark config must use the namespace IPs. Use a large iteration
+count because current `daqiri_bench_socket` treats `iterations: 0` as zero work,
+not as "run until --seconds expires".
```
dleshchev, line 216: probably need an issue for that

src/managers/socket/daqiri_socket_mgr.cpp
```diff
@@ -1340,6 +1347,16 @@ Status SocketMgr::socket_connect_to_server(const std::string& dst_addr, uint16_t
...
 
     if (cfg_.common_.protocol == SocketProtocol::TCP) {
+      if (ep->primary_conn_id != 0 && ep->socket_cfg.remote_ip_ == dst_addr &&
+          ep->socket_cfg.remote_port_ == dst_port &&
+          (src_addr.empty() || src_addr == ep->socket_cfg.local_ip_)) {
+        std::lock_guard<std::mutex> lock(state_mutex_);
+        if (connections_.find(ep->primary_conn_id) != connections_.end()) {
+          *conn_id = ep->primary_conn_id;
+          return Status::SUCCESS;
+        }
```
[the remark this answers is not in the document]
dleshchev: Seems like closing connections is still needed?
after bouncing it with claude:
The running guards make reuse correct but not bounded. Under TCP server connection churn, three things grow forever and are only reclaimed at shutdown():

1. connections_ — dead server entries never erased.
2. server_connections_[key] — a push-only vector (:1166-1167, :1211-1212), never trimmed; grows one entry per accepted connection and accumulates dead IDs that the loop above must skip.
3. Finished-but-unjoined std::thread rx_thread inside each retained ConnectionState — not joined until close_all_connections(). Accumulating joinable threads leaks pthread handles and eventually breaks std::thread creation on a long-lived server.

dleshchev (commented):
looks like dead connections can be still an issue; also the .md file needs to be placed somewhere more appropriate.

SPARK_SOCKET_WIRE_README.md
cliffburdick: This tutorial is specific to spark if that's what you mean

cliffburdick:
Added into PR #114

PR #111, opened 2026-06-01
#112 - Add OpenTelemetry Metrics

cliffburdick:
This PR adds OpenTelementry-compatible metrics for use in exporting to libraries like Prometheus and Grafana. A working example using the raw_tx_rx is provided.


dleshchev:
While running the Grafana example, we noticed the DPDK per-interface Prometheus series were staying at zero even though the queue-level series were moving. Specifically, the `queue_id="all"` counters for `daqiri_rx_bytes_total`, `daqiri_tx_bytes_total`, `daqiri_rx_packets_total`, and `daqiri_tx_packets_total` were not populated when queue xstats were available, while `queue_id="0"` had live values.

I pushed commit `134ad5e` to always populate the DPDK `queue_id="all"` packet/byte metrics from `rte_eth_stats_get()`, while keeping queue-specific metrics sourced from queue xstats. Rebuilt and reran the dashboard stack; the `all` RX/TX byte counters are now nonzero and track the port-level traffic.

PR #115, opened 2026-06-02
#113 - Fix RDMA cross-host SEND timeout and shutdown deadlock

### LINKED ISSUE #113: [BUG] RDMA cross-host SEND immediately fails with transport_retry_exceeded and the server never posts receives
RDMA between two DGX Sparks back-to-back dies on the first SEND with `IBV_WC_RETRY_EXC_ERR`. The RDMA-CM handshake finishes fine (both sides hit `RDMA_CM_EVENT_ESTABLISHED`), but right after the server creates its listener it logs `[CRITICAL] Couldn't find server params for address 169.254.95.47` — so whatever posts the receive WRs after establish never runs, and the client's first SEND has nothing to land in.

Single-host RDMA works on the same build, and `ib_send_bw` over the same cable runs at ~212 Gb/s, so I don't think it's the wire.

Looks like an address-key mismatch between where `server_params` gets inserted (listener-create path around `daqiri_rdma_mgr.cpp:1056`) and where it gets looked up (`:692`) — probably string-vs-binary or `INADDR_ANY`-vs-explicit.

Two Sparks back-to-back on ConnectX-7 p0, server on `169.254.95.47`, client on `169.254.100.253`. Build per `docs/tutorials/bare-metal-cmake-build.md`. Start the server first:

```bash
# server (spark-201a)
sudo build/.../daqiri_bench_rdma examples/daqiri_bench_rdma_server_spark_xhost.yaml --mode server --seconds 30

# client (spark-960b)
sudo build/.../daqiri_bench_rdma examples/daqiri_bench_rdma_client_spark_xhost.yaml --mode client --seconds 30
```

Client tail:

```
[ERROR] daqiri_rdma_mgr.cpp:489: CQ error on client: transport retry counter exceeded (12) for WRID 4660 ...
Client received messages: 0
```

Server tail (one line after the listener gets created at `:1056`):

```
[CRITICAL] daqiri_rdma_mgr.cpp:692: Couldn't find server params for address 169.254.95.47
[CRITICAL] daqiri_rdma_mgr.cpp:718: Couldn't find an available queue ID for server 169.254.95.47:4096 ...
```

Full logs attached: `logs/spark_rdma_tx.log`, `logs/spark_rdma_rx.log`.

**Expected behavior**

Server posts receives at `RDMA_CM_EVENT_ESTABLISHED`, client SENDs land, both sides report non-zero throughput, no `RETRY_EXC_ERR`. Single-host `daqiri_bench_rdma_tx_rx_spark.yaml` keeps working.

**Environment overview**

- Environment location: Bare-metal
- Method of DAQIRI install: source, branch `docs/bare-metal-cmake-build`, configured with `-DDAQIRI_MGR="dpdk socket rdma"`

**Environment details**

- OS: IGX OS / Ubuntu 24.04 ARM (aarch64), kernel 6.14
- Hardware: 2x DGX Spark (GB10), ConnectX-7 fw `28.45.4028`, back-to-back on `enp1s0f0np0`
- CUDA 13.0, driver 580.95.05, DPDK 25.11 patched with `dmabuf.patch` + `dpdk.nvidia.patch`, `libibverbs-dev`/`librdmacm-dev` 2510.0.11-1
- `nvidia-peermem` not loaded (dma-buf path); YAMLs use `kind: host_pinned`

**Additional context**

Raw GPUDirect cross-host works on the same hosts/YAMLs, so this is RDMA-specific. There's also a server-side segfault that shows up on stderr but didn't make it into my log capture — it seems like it's downstream of the missing receive-provisioning state, not a separate bug, but worth confirming once the lookup is fixed.

chloecrozier:
## Summary

Closes #113 (RDMA cross-host SEND fails with `IBV_WC_RETRY_EXC_ERR`) plus a graceful-shutdown deadlock the #113 fix exposes, and adds cross-host DGX Spark example YAMLs as the regression test.

## Fixes (`src/managers/rdma/daqiri_rdma_mgr.{h,cpp}`)

1. **Wrong interface index on the server (#113).** When a client connects, the server picks which configured interface to use for that connection. It was using the wrong number: the hardware's port number from the InfiniBand driver, instead of the index into our own configured interface list. The two happen to look similar but mean different things, so the server thread immediately crashed trying to look up an interface that wasn't there. Because the server crashed before posting any receives, the client's first send had nothing to land in and timed out with `RETRY_EXC_ERR`. Fix: use the interface index we already saved when the listener was set up.

2. **Server hangs on graceful shutdown (only reachable once the first issue was fixed).** When the client disconnects, the server tries to clean up its worker thread. It was doing this badly in three ways at once: it held a lock while waiting for the worker, the worker had no way to know it should stop (it only watched the global "everything is shutting down" flag), and the stats-printing code also wanted that same lock. Once issue 1 stopped the server from crashing early, every clean run ended up stuck in this three-way deadlock and had to be `SIGKILL`ed. Fix: give each worker its own "please exit" flag, set it when the client disconnects, and release the lock before waiting for the worker to actually stop.

The `Couldn't find server params for address …` line that may show up once at startup is a benign timing race (the app polls slightly before the listener is registered), not the bug.

## Regression test

Three new `examples/` YAMLs exercising the cross-host data plane that no existing config reached:

- `daqiri_bench_raw_{tx,rx}_spark_xhost.yaml`: split per-host (raw bench has no `--mode`).
- `daqiri_bench_rdma_tx_rx_spark_xhost.yaml`: single config selected via `--mode {client,server}`, matching the existing one-file-per-workload pattern.

Reuses the `1.1.1.1` / `2.2.2.2` IPs and CPU pinning of the existing `_spark.yaml` configs, with each address bound to its host's p0 instead of both on one box.

## Tests ran to verify

- [x] `python scripts/check_doc_refs.py`: 23 YAMLs, all covered.
- [x] `mkdocs build --strict`: clean.
- [x] `python scripts/check_html_links.py site/`: clean.

Hardware (two cross-cabled DGX Sparks, GB10, ConnectX-7 fw `28.45.4028`, IGX OS / Ubuntu 24.04 ARM, kernel 6.14, CUDA 13.0, driver 580.95.05):

- [x] **RDMA xhost**: `Server received: 21761`, `Client received: 15256` (was `0` / `0` pre-fix); zero `transport retry counter exceeded`; server reaches `Removed leftover hugepage file …` and exits ~3 s after its `--seconds` timer with no `SIGKILL`. Logs: `logs/spark_rdma_{rx,tx}_post_fix3.log` (vs. pre-fix `logs/spark_rdma_{rx,tx}.log`).
- [x] **Raw GPUDirect xhost**: lossless over 15 s. TX `24,657,920` / RX `24,657,920` / dropped `0` (~106 Gbps). Logs: `logs/spark_raw_{rx,tx}_post_fix.log`.
- [x] **Single-host regression (`daqiri_bench_rdma_tx_rx_spark.yaml`)**: same code paths, unchanged behavior.


chloecrozier:
### Addressed Greptile's review:
ready_to_exit is now std::atomic<bool> (matching the existing rdma_force_quit pattern, so the worker actually observes the write under the C++ memory model), and the per-connection rings are returned to the pool only after worker_to_join.join() so the worker can't be observed touching a "freed" ring.

Issue #119, opened 2026-06-03
[FEA] Add uninstall script for the CMake install and container image


chloecrozier:
## Summary + Solution

Add `scripts/uninstall.sh` so a developer can remove DAQIRI's CMake install or container image without touching the build prerequisites (DPDK, DOCA libraries, CUDA, hugepages, NIC drivers). Targets are `cmake`, `container`, and `all`; the script supports `--dry-run`, `--yes`, and a built-in verification step. Short usage notes are added inline to `docs/tutorials/bare-metal-cmake-build.md` (new Step 7) and `docs/getting-started.md` with no new tutorial page or `mkdocs.yml` change

## Test plan after solution is implemented

- Dry-run both targets on a host with a working install and confirm the printed actions match `build/install_manifest.txt`
- Real run to confirm the verification block reports `OK` for every check

PR #120, opened 2026-06-04
#119 - Add uninstall script for the CMake install and container image

chloecrozier:
## Summary
Add `scripts/cleanup.sh` to remove DAQIRI's CMake install or container image without touching build prerequisites (DPDK, DOCA libraries, CUDA, hugepages, NIC drivers). Targets are `cmake`, `container`, and `all`, with `--dry-run`, `--yes`, and built-in verification. The cmake target is manifest-driven with a name-scoped fallback scan; container verification surfaces (without auto-removing) dangling images and exited containers from interrupted builds. Short usage notes are added to `docs/tutorials/bare-metal-cmake-build.md` and `docs/getting-started.md`.

## Test plan
- Dry-run both targets on a host with a working install and confirm the printed actions match `build/install_manifest.txt`.
- Real run; confirm the verification block reports `OK` for every check.

## To Do:
- [x] IGX
- [x] DGX Spark
- [x] RTX Pro 6000


chloecrozier:
Addressed and applied greptile's comments.

RamyaGuru:
I tested uninstall.sh container on my IGX. The dry run and full unisntall behavior all looked good! Maybe @dleshchev can test the cmake uninstall on his system before we merge this in.

One minor semantic thing: I think we discussed yesterday calling this script something like "cleanup.sh" instead of "uninstall.sh".

docs/tutorials/bare-metal-cmake-build.md
```diff
@@ -396,6 +396,20 @@ A successful run prints a stream of `[INFO]` lines followed by an RX/TX rate sum
 
     The user-space libraries are installed but the kernel drivers are not; `lsmod | grep mlx5` will be empty. Follow [Check your NIC drivers](system_configuration.md#check-your-nic-drivers) to install `doca-ofed` and reboot.
 
+## Step 7: Uninstall
+
+To remove the CMake install while keeping every prerequisite (DPDK, DOCA libraries, CUDA, hugepages, NIC drivers) in place, run [`scripts/uninstall.sh`](https://github.com/NVIDIA/daqiri/blob/main/scripts/uninstall.sh) with the `cmake` target:
```
RamyaGuru: Instead of making this "Step 7" of the tutorial, why don't we make Uninstall instructions just a separate note at the bottom

chloecrozier: Fixed!

Issue #54, opened 2026-05-06
[DOC] Update system tuning tutorial with RTX PRO 6000 specific details


chloecrozier:
## Report needed documentation

Need RTX PRO 6000 specific details added to the system tuning tutorial. The current tutorial covers IGX Orin and DGX Spark (added in #49), which are different from a workstation/server with an RTX PRO 6000 and the users need to be informed about the differences.

PR #121, opened 2026-06-04
#54 - Add RTX PRO 6000 Blackwell Server Edition support to tune_system.py

chloecrozier:
Updates `python/tune_system.py` for RTX PRO 6000 Blackwell Server Edition alongside the existing IGX / DGX Spark paths. Every check still answers "is the system tuned for max throughput?".

## Added

- Detects RTX PRO 6000 Blackwell Server Edition cards
- Flags low BAR1 only on that card
- Explains why peermem is optional on Blackwell

## Fixed

- `ibdev2netdev` missing no longer crashes `--check all`
- CPU governor check logs one summary line per distinct governor (not per CPU)
- `ibdev2netdev` warning emits once per run, not three times

## Example output (dev box: 5× RTX PRO 6000 Blackwell SE, 256-core EPYC)

```text
$ sudo python3 python/tune_system.py --check cpu-freq
ERROR - CPU governor: scaling_governor file not found on 256/256 online CPUs.
        The cpufreq driver may not be loaded (e.g. amd-pstate, intel_pstate, or
        cppc_cpufreq). Performance scaling cannot be checked.

$ sudo python3 python/tune_system.py --check bar1-size
INFO - GPU 00000000:04:00.0: BAR1 size is 131072 MiB.
INFO - GPU 00000000:73:00.0: BAR1 size is 131072 MiB.
INFO - GPU 00000000:74:00.0: BAR1 size is 131072 MiB.
INFO - GPU 00000000:84:00.0: BAR1 size is 131072 MiB.
INFO - GPU 00000000:F3:00.0: BAR1 size is 131072 MiB.

$ sudo python3 python/tune_system.py --check peermem
INFO - nvidia-peermem module is not loaded. On this RTX PRO 6000 Blackwell Server
       Edition system /dev/dma_heap/system is available, so the patched DPDK shipped
       with this repo (dpdk_patches/dmabuf.patch) takes the dma-buf GPUDirect path
       and does not need peermem. If you are building DAQIRI against stock DPDK
       instead, load nvidia-peermem.

$ sudo python3 python/tune_system.py --check mrrs   # when ibdev2netdev is missing
WARNING - The ibdev2netdev command is not found (try: apt install infiniband-diags).
          Skipping NIC-dependent checks (mrrs, mps, mtu).
```

### DIFF
--- a/python/tune_system.py
+++ b/python/tune_system.py
@@ -44,6 +44,7 @@ PCIE_EFFECTIVE_GB_PER_SEC_PER_LANE = {
     32.0: 3.938,
     64.0: 7.563,
 }
+BAR1_BLACKWELL_MIN_MIB = 32768  # 32 GiB
 
 
 @dataclass
@@ -204,30 +205,6 @@ def _dmabuf_gpu_path_available():
     return os.path.exists("/dev/dma_heap/system")
 
 
-def _gpu_name_by_bdf():
-    """
-    Returns {pci_bdf: product_name} for every visible NVIDIA GPU, or {} if
-    nvidia-smi is unavailable. Used by per-GPU checks (e.g. check_bar1_size)
-    to apply Blackwell-specific thresholds only to the Blackwell GPU(s) in a
-    heterogeneous system rather than to every GPU in the box.
-    """
-    try:
-        result = subprocess.run(
-            ["nvidia-smi", "--query-gpu=pci.bus_id,name", "--format=csv,noheader"],
-            capture_output=True,
-            text=True,
-            check=True,
-        )
-    except (FileNotFoundError, subprocess.CalledProcessError):
-        return {}
-    names = {}
-    for line in result.stdout.splitlines():
-        parts = line.split(",", 1)
-        if len(parts) == 2:
-            names[parts[0].strip()] = parts[1].strip()
-    return names
-
-
 def check_peermem_kernel():
     """
     Check if the nvidia-peermem module for GPUDirect is loaded in the kernel.
@@ -377,8 +354,9 @@ def get_online_cpus():
 def check_cpu_governor():
     """
     Checks if the CPU frequency governor is set to 'performance' for all online CPUs.
-    Output is bucketed by result so a 256-core system does not emit 256 lines when
-    every CPU is in the same state. Per-CPU detail still surfaces if results vary.
+    Aggregates results by governor value and logs one summary line per distinct
+    governor (e.g. 256/256 online CPUs set to 'performance'), plus separate counts
+    for CPUs whose scaling_governor file is missing or unreadable.
     """
     online_cpus = get_online_cpus()
     total = len(online_cpus)
@@ -684,11 +662,10 @@ def check_bar1_size():
     # 32 GiB is the conservative "rebar is fully unlocked" floor: well below
     # the 96 GB card capacity but high enough that any platform exposing less
     # is almost certainly missing Resizable BAR / Above 4G Decoding in BIOS.
-    # The threshold is applied per-GPU via gpu_names below so heterogeneous
+    # The threshold is applied per-GPU via gpu_info below so heterogeneous
     # boxes (e.g. RTX PRO 6000 + H100) only get the Blackwell rule on the
     # Blackwell card.
-    BAR1_BLACKWELL_MIN_MIB = 32768  # 32 GiB
-    gpu_names = _gpu_name_by_bdf()
+    gpu_info_by_bdf = get_nvidia_gpu_info_by_bdf()
     try:
         # Run nvidia-smi to get BAR1 memory information
         result = subprocess.run(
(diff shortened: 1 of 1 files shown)


chloecrozier:
Addressed and applied the greptile suggestions

examples/CMakeLists.txt
```diff
@@ -9,9 +9,17 @@ project(daqiri_bench_examples LANGUAGES CXX CUDA)
...
-if(DAQIRI_EXAMPLES_USING_DEFAULT_CUDA_ARCHITECTURES AND
-   CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 13.0)
-  list(APPEND CMAKE_CUDA_ARCHITECTURES 121)
+if(DAQIRI_EXAMPLES_USING_DEFAULT_CUDA_ARCHITECTURES)
+  if(CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 13.0)
+    list(APPEND CMAKE_CUDA_ARCHITECTURES 121)
+  else()
+    message(WARNING
+      "CUDA Toolkit ${CMAKE_CUDA_COMPILER_VERSION} is older than 13.0; sm_120 "
+      "(RTX PRO 6000 Blackwell Server Edition) will be omitted from "
```
cliffburdick: What's the reason we have a message specific to this GPU? In theory we should support any reasonably new GPU I think

python/tune_system.py
```diff
@@ -214,6 +290,14 @@ def check_peermem_kernel():
                 "(e.g. GB10 / DGX Spark) where peermem does not apply. Use kind: host_pinned "
                 "in the daqiri YAML for GPUDirect on this platform."
             )
+        elif is_any_blackwell_server_edition_discrete() and _dmabuf_gpu_path_available():
```
cliffburdick: Same question as above about having a message specific to a GPU type

examples/CMakeLists.txt
chloecrozier: Good point, I removed it! I see that the tune script already catches arch mismatches at runtime, so this is redundant

python/tune_system.py
chloecrozier: Yeah since dma-buf availability is what matters, I dropped the SKU gate and just made the message generic

python/tune_system.py
```diff
@@ -193,6 +194,40 @@ def is_any_integrated_gpu():
...
+    """
+    Returns True if the kernel exposes the dma-buf path that recent NVIDIA drivers
+    use for GPUDirect in place of nvidia-peermem. The patched DPDK shipped with
+    this repo (dpdk_patches/dmabuf.patch) takes this path on platforms that
+    expose it, which is why peermem is not required there.
+    """
+    return os.path.exists("/dev/dma_heap/system")
+
+
+def _gpu_name_by_bdf():
```
dleshchev: doesn't it mirror existing get_nvidia_gpu_info_by_bdf?

python/tune_system.py
```diff
@@ -328,30 +377,49 @@ def get_online_cpus():
 def check_cpu_governor():
     """
     Checks if the CPU frequency governor is set to 'performance' for all online CPUs.
+    Output is bucketed by result so a 256-core system does not emit 256 lines when
+    every CPU is in the same state. Per-CPU detail still surfaces if results vary.
```
dleshchev: not sure if this is correct - so far it only reports overall stats

python/tune_system.py
```diff
@@ -610,6 +678,17 @@ def check_bar1_size():
...
+    # On RTX PRO 6000 Blackwell Server Edition (96 GB GDDR7) the generic
+    # > 1024 MiB threshold passes trivially even with Resizable BAR disabled
+    # (the card still exposes a multi-GiB BAR1 in some platform configs).
+    # 32 GiB is the conservative "rebar is fully unlocked" floor: well below
+    # the 96 GB card capacity but high enough that any platform exposing less
+    # is almost certainly missing Resizable BAR / Above 4G Decoding in BIOS.
+    # The threshold is applied per-GPU via gpu_names below so heterogeneous
+    # boxes (e.g. RTX PRO 6000 + H100) only get the Blackwell rule on the
+    # Blackwell card.
+    BAR1_BLACKWELL_MIN_MIB = 32768  # 32 GiB
```
dleshchev: do we want to move it to the top along with other constants?

dleshchev (approved):
Can be merged after fixing some nits below and one more thing (id'ed by claude)

The PR body lists under Added: "CMake warns when CUDA Toolkit < 13.0 omits sm_120." That is not in this diff (git show HEAD --name-only → only tune_system.py), and the only related in-tree logic doesn't match the claim on two counts:

- CMakeLists.txt:28-31 silently appends arch 121 when CUDA ≥ 13.0 — it does not warn when CUDA < 13.0.
- Arch 121 is GB10 / DGX Spark (sm_121), per AGENTS.md. RTX PRO 6000 Blackwell SE is sm_120, which is not in the default arch list (80;90 + 121).

Net effect: the tuning script now advertises/validates RTX PRO 6000 support, but a from-source build with default CMAKE_CUDA_ARCHITECTURES won't actually compile sm_120 kernels for it. Please either (a) drop the CMake bullet from the description if it belongs to a different PR, or (b) include the intended CMake change here — and if real RTX PRO 6000 support is the goal, confirm whether 120 needs adding to the default arch list (out of scope for this file, but it's what "support" implies). At minimum the description should match what merges.

chloecrozier:
@dleshchev, I removed the duplicate GPU helper and reused the existing one, corrected the cpu-freq docstring, and moved the BAR1 constant to the top with the others. For RTX PRO 6000 builds, users still need to pass -DCMAKE_CUDA_ARCHITECTURES=120 manually, but that is already covered in doc. I also cleaned up some duplicate nvidia-smi calls

PR #132, opened 2026-06-08
#131 - Fix DPDK TX enqueue mbuf leak

### LINKED ISSUE #131: [BUG] DPDK TX enqueue failure leaks packet mbufs
**Describe the bug**
`DpdkMgr::send_tx_burst()` leaks packet mbufs when `rte_ring_enqueue()` fails. The failure path calls `free_tx_burst(burst)`, which returns the TX burst pointer arrays and metadata, but it does not free the packet mbufs allocated by `get_tx_packet_burst()`.

1. Use the DPDK raw TX path.
2. Allocate a TX burst and packet mbufs with `get_tx_packet_burst()`.
3. Populate the burst.
4. Call `send_tx_burst()` when the TX work ring is full, causing `rte_ring_enqueue()` to fail.

At that point, `send_tx_burst()` returns `NO_SPACE_AVAILABLE` after freeing only the burst metadata/pointer arrays, so the application no longer has handles for the packet mbufs.

**Expected behavior**
If `send_tx_burst()` cannot enqueue the burst, DAQIRI should free all unsubmitted packet mbufs before releasing the burst metadata.

 - Environment location: Source review
 - Method of DAQIRI install: source

**Environment details**
 - DAQIRI version: current `main`

**Additional context**
The public API docs distinguish `free_tx_burst()` from `free_all_packets_and_burst_tx()`: `free_tx_burst()` frees metadata only and expects packet buffers to have already been freed. On the enqueue-failure path, the packets were never submitted to the TX worker, so `send_tx_burst()` must clean them up itself.

fallintoplace:
## Summary
- Free unsubmitted DPDK TX packet mbufs when the TX work ring enqueue fails.
- Keep the existing burst metadata/pointer-array cleanup after packet buffers are released.

## Why
`free_tx_burst()` only returns TX burst metadata and pointer arrays. If `send_tx_burst()` hits a full ring after packet mbufs were allocated, releasing only the burst leaves the application without handles to free those mbufs.

For multi-segment TX bursts, the worker chains mbufs only after enqueue succeeds, so the failure path frees each allocated segment directly.

## Validation
- `git diff --check`
- `IMAGE_TAG=daqiri:tx-enqueue-leak BASE_TARGET=dpdk DAQIRI_MGR=dpdk DAQIRI_BUILD_PYTHON=OFF BUILD_SHARED_LIBS=ON scripts/build-container.sh`

Fixes #131

### DIFF
--- a/docs/api-reference/cpp.md
+++ b/docs/api-reference/cpp.md
@@ -241,6 +241,13 @@ daqiri::send_tx_burst(burst);
 
 The burst is enqueued to the TX worker thread, which sends it to the NIC via DMA.
 
+`send_tx_burst()` takes ownership of the burst on success and on a full-ring
+failure: on `SUCCESS` the TX worker owns it, and on `NO_SPACE_AVAILABLE` (the TX
+ring is full) it has already freed the packets and the burst internally. In both
+cases the application must **not** free or otherwise access the burst afterwards.
+`NO_SPACE_AVAILABLE` is the only failure a correctly-configured sender encounters
+at runtime.
+
 ### Timed Transmission
 
 For precise packet scheduling (requires ConnectX-7+):
--- a/docs/api-reference/python.md
+++ b/docs/api-reference/python.md
@@ -336,12 +336,15 @@ for idx in range(daqiri.get_num_packets(burst)):
 
 ```python
 status = daqiri.send_tx_burst(burst)
-if status != daqiri.Status.SUCCESS:
-    daqiri.free_all_packets_and_burst_tx(burst)
+# Do not free `burst` after a SUCCESS or NO_SPACE_AVAILABLE return — see the ownership note below.
 ```
 
-After a successful `send_tx_burst()`, the worker thread owns the burst and the
-application must not access it again.
+`send_tx_burst()` takes ownership of the burst on success and on a full-ring
+failure: on `SUCCESS` the worker thread owns it, and on `NO_SPACE_AVAILABLE` (the
+TX ring is full) it has already freed the packets and the burst internally. In
+both cases the application must **not** free or otherwise access the burst
+afterwards. `NO_SPACE_AVAILABLE` is the only failure a correctly-configured
+sender encounters at runtime.
 
 ### Timed Transmission
 
@@ -396,9 +399,9 @@ if status == daqiri.Status.SUCCESS:
     if daqiri.get_tx_packet_burst(tx_burst) == daqiri.Status.SUCCESS:
         daqiri.copy_buffer_to_packet(tx_burst, 0, b"hello")
         daqiri.set_packet_lengths(tx_burst, 0, [5])
-        status = daqiri.send_tx_burst(tx_burst)
-        if status != daqiri.Status.SUCCESS:
-            daqiri.free_all_packets_and_burst_tx(tx_burst)
+        # send_tx_burst() takes ownership of tx_burst on success and on a
+        # full-ring failure, so do not free it here.
+        daqiri.send_tx_burst(tx_burst)
     else:
         daqiri.free_tx_metadata(tx_burst)
 ```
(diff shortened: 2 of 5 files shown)


dleshchev:
Re: Greptile's P1 — **mbuf leak on the `INVALID_PARAMETER` early return** (`src/managers/dpdk/daqiri_dpdk_mgr.cpp:4461-4465`).

Confirmed and tracked in #164 (it was independently flagged by two other reviewers as well). Deferring it out of this PR to keep the scope to the ring-full leak #131 targets, for two reasons:

1. **The suggested inline fix would crash, not just leak.** Applying the same `free_unsubmitted_tx_packets(burst)` + `free_tx_burst(burst)` pattern here isn't safe: this path is reached precisely because the port/queue key is missing from `tx_rings`, and that same key is therefore missing from `tx_burst_buffers`. `free_tx_burst()` (`:4402-4408`) does an **unchecked** `tx_burst_buffers.find()` and dereferences `burst_pool->second`, so calling it on a bad-key burst is UB. A correct fix has to harden that lookup first (and even then can't return the pointer-array buffers without the burst carrying its own pool handle).

2. **Near-unreachable for well-formed bursts.** A burst handed to `send_tx_burst()` came from `get_tx_packet_burst()`, which already validated the same key against `tx_dpdk_q_map_`/`tx_burst_buffers`. So a valid burst never hits this `INVALID_PARAMETER` return; it only triggers on a hand-constructed/misconfigured burst or inconsistent internal map state.

#164 captures the full resolution (harden `free_tx_burst` → free on the early-return paths in DPDK + socket → widen the ownership docs back to a simple "never free after `send_tx_burst()`"). The docs in this PR are intentionally scoped to "consumes on `SUCCESS` and `NO_SPACE_AVAILABLE`" until that lands.

dleshchev:
Update: rather than defer, I've folded the fix into this PR (commit 533d081), so the Greptile P1 is now resolved and `send_tx_burst()` consumes the burst on **every** return path.

- **DPDK** (`daqiri_dpdk_mgr.cpp`): `free_tx_burst()` now uses a checked `tx_burst_buffers` lookup (the previous unchecked `find()` dereferenced `end()` on a missing key — the crash hazard I flagged earlier), plus a null-burst guard. The `INVALID_PARAMETER` early return now frees via `free_unsubmitted_tx_packets()` + `free_tx_burst()`.
- **Socket** (`daqiri_socket_mgr.cpp`): the UDP/TCP malformed-burst `INVALID_PARAMETER` path now frees, mirroring the existing no-endpoint cleanup.
- **RoCE: unchanged.** It delegates to the RDMA manager, which already frees on its failure paths; the roce-not-initialized early return never owns an allocated burst (`get_tx_packet_burst` also returns there without allocating), so there's nothing to free.
- **docs**: ownership note widened back to "every return path."

Verified with a full incremental build (`daqiri_dpdk`, `daqiri_socket`, `daqiri_rdma`, all examples) + clang-format clean. This supersedes my earlier "deferring to #164" comment.

dleshchev:
**Reverted the fold** (force-pushed back to da08606). On review, freeing on the `INVALID_PARAMETER` path is unsafe: TX metadata from `tx_metadata` is recycled without reset (no obj-init at `daqiri_dpdk_mgr.cpp:2542`; `free_tx_burst` never nulls `pkts[]`/resets `num_segs`), and that path is reached exactly when `get_tx_packet_burst()` hasn't populated the burst — so `free_unsubmitted_tx_packets()` would free stale `pkts[]` it doesn't own (corruption, worse than the leak it was closing).

So this PR is back to the **narrow, verified contract**: `send_tx_burst()` consumes the burst on `SUCCESS` and `NO_SPACE_AVAILABLE`; docs/example scoped to match. That's the safe, in-scope fix for #131.

The `INVALID_PARAMETER` cleanup (the Greptile P1) needs a TX-metadata-reset prerequisite and stays tracked in #164 with the full rationale. #163 (multi-segment `get_tx_packet_burst` partial-alloc leak) remains a separate follow-up. Apologies for the churn — the fold was an overcorrection.

PR #98, opened 2026-05-26
#15 - Add DGX Spark v1 performance report

RamyaGuru:
## Summary

Rebuilt on current `main`. Adds `docs/performance-dgx-spark.md` with the DGX Spark performance report, performance-matrix CSS, landing-page/nav/README links, and the docs-sync update for per-platform performance docs.

## Current status

This PR intentionally carries forward the existing DPDK/RoCE measurements. The RoCE small-payload cells are marked pending a current-main hardware refresh after the RDMA bench changes. Socket UDP/TCP rows remain deferred until the separate Spark rerun fills them.

## Test plan

- [x] `git diff --check origin/main..HEAD`
- [x] `python3 scripts/check_doc_refs.py`
- [x] `.venv-docs/bin/mkdocs build --strict --site-dir /tmp/daqiri-pr98-site`

PR #138, opened 2026-06-09
 #126 - Fix intermittent RoCE small-message throughput cliff (RNR timer)

### LINKED ISSUE #126: [BUG]  RDMA bench: intermittent throughput cliff at small messages / low pace targets
## Summary

The RoCE/RDMA path (`daqiri_bench_rdma` + `libdaqiri_rdma`) intermittently collapses
from line rate (~97 Gb/s) to near-zero (~0.001–0.7 Gb/s) on a true over-the-wire
loopback. The collapse is **not** a clean function of message size — it is a race:
the same configuration runs healthy in some cells and cliffs in others. Large
messages (1–8 MB) are usually fine; sub-1 MB messages cliff almost every time, and
even an 8 MB run can cliff at a specific pace target. Throughput collapses to ~11–230
ops/s with **zero CQE errors and zero drops**, so the QP/wire are healthy — the stall
is software-side, on the bench/manager op-rate path.

This is the issue previously tracked internally as the "RDMA pps cliff" (follow-up #5).

## Environment

- DGX Spark (GB10), single ConnectX-7, two cabled ports in a netns wire loopback
  (`dq_wire_client` 10.250.0.1 / `dq_wire_server` 10.250.0.2, RoCEv2, RC).
- `daqiri_bench_rdma` run split-process: `--mode server` and `--mode client`, each in
  its own network namespace (see `scripts/setup_spark_wire_loopback_netns.sh`,
  `examples/daqiri_bench_rdma_tx_rx_spark_netns_{server,client}.yaml`).
- MRs `kind: host_pinned`, `num_bufs: 20`. Driver: `examples/run_spark_bench.sh rdma`.
- Branch `15-bench-spark-perf-doc`, library built **after** #115 (cross-host SEND
  timeout + shutdown deadlock fix). Crash/0-byte symptoms from the stale pre-#115
  library are a separate, already-resolved problem — this is observed on the fixed lib.

## Reproduction

```bash
# bring up the netns wire loopback (in the privileged container, as root)
./scripts/setup_spark_wire_loopback_netns.sh up && ./scripts/setup_spark_wire_loopback_netns.sh verify

# message-size sweep (batch 1, unpaced) -- sub-1 MB cells cliff
./examples/run_spark_bench.sh rdma sweep

# pace sweep at 8 MB -- an intermittent target collapses
./examples/run_spark_bench.sh rdma drop-curve
```

## Evidence (2026-06-05)

**Message-size sweep** (`bench-results/20260605T194748Z-rdma-sweep`, batch 1, unpaced):

| message_size | gbps  | pps    | drops |
|--------------|-------|--------|-------|
| 8 MB         | 97.49 | 1,523  | 0     |
| 1 MB         | 96.92 | 11,553 | 0     |
| 64 KB        | 0.121 | 231    | 0     |
| 8 KB         | 0.005 | 74     | 0     |
| 4 KB         | 0.001 | 33     | 0     |

pps *falls* as messages shrink (it should rise) — i.e. a per-op stall, not an
efficiency loss.

**Pace sweep at 8 MB** (`bench-results/20260605T195211Z-rdma-drop-curve`):

| target gbps | achieved | drops |
|-------------|----------|-------|
| 1 / 5 / 10  | 0.99 / 4.97 / 9.97 | 0 |
| **25**      | **0.695** (cliff) | 0 |
| 50 / 75 / 100 | 49.98 / 74.98 / 96.40 | 0 |
| unpaced     | 96.89 | 0 |

`target=25` collapsed while its neighbours were perfect — confirms the cliff is
**intermittent / a race**, not deterministic by shape.

## Ruled out (as the *explanation for the cliff variance*)

- **CQE errors / drops** — zero in every affected cell; the QP and wire are healthy.
- **`get_rx_burst` "No server RX ring found" criticals as a *predictor*** — the count is
  ~26,000×/run in **every** cell, healthy and cliffed alike (e.g. 26,037 in the 96.4 Gb/s
  cell vs 25,968 in the 0.695 Gb/s cell). Because the count does **not** vary with
  throughput, it does not *explain why one cell cliffs and its neighbour doesn't*.

  **Important:** this only rules the spam out as a *predictor of the per-cell variance* —
  it does **not** mean the spam is harmless or unrelated. The underlying condition (the
  server's RX ring lookup returning null in steady state) is itself wrong and may share a
  root cause with the cliff (see below).

## The error spam is itself a bug (not valid behavior)

`get_rx_burst` emits `CRITICAL` ~866×/second in steady state on a healthy run. A
hot-path `CRITICAL` is wrong by definition — that level means "something is broken" — and
it has two concrete problems:

1. **Wrong lookup.** `rx_rings_map_[conn_id]` (server branch
   `src/managers/rdma/daqiri_rdma_mgr.cpp:1777`, client branch `:1787`) uses `operator[]`,
   which **inserts a null entry on a miss** instead of reporting absence. It should be
   `.find()`. The null-insert is also actively harmful: it can **poison
   `rdma_get_server_conn_id`'s gate at `:854`**, which checks
   `rx_rings_map_.find(client_id) == end()`. Once `operator[]` has inserted a null-valued
   entry, that key now *exists*, so the gate passes and hands back a conn_id whose ring is
   null → guaranteed miss on the next `get_rx_burst`, repeated every drain attempt.
2. **Wrong level / control flow.** Returning `INVALID_PARAMETER` makes the bench's drain
   loop `break` immediately (`examples/rdma_bench.cpp:133`), so the server stops draining
   RX completions for that iteration. If a missing/absent ring is ever a legitimate
   transient (polled before the ring is wired up, or a role with no RX ring), the correct
   response is a quiet `NOT_READY`, not a `CRITICAL` + `INVALID_PARAMETER`. At minimum the
   log should be `DEBUG` and/or rate-limited.

## Suspected area

The collapse looks like a credit/completion-draining stall in the bench worker loop
(`examples/rdma_bench.cpp:130-166`) and/or the manager's RX completion path
(`src/managers/rdma/daqiri_rdma_mgr.cpp`), where the server worker fails to drain
RX completions for a connection and stops returning credits, throttling the op-rate.
The race-like dependence on timing (per-cell, per-target) points at the
connection/ring lifecycle across the CM thread, worker threads, and `threads_mutex_`
(`rdma_get_server_conn_id` gating at ~`daqiri_rdma_mgr.cpp:854`; ring populate at
`:375`, erase at `:1417`).

**These two symptoms — the steady-state null RX ring and the intermittent cliff — may
share one root cause.** A server that cannot reliably resolve its RX ring to drain
completions is exactly the kind of thing that stalls the op-rate; the constant ~26k count
only tells us the *spam frequency* doesn't vary with throughput, not that the underlying
null-ring condition is unrelated to the stall. The investigation should start by
explaining **why `rx_rings_map_` yields a null ring at all** (is `qp_params.rx_ring`
null when stored at `:375`? is the entry erased/re-created mid-run at `:1417`? is the
poisoned-`operator[]` gate at `:854` handing back a dead conn_id?) and only then decide
whether the cliff is the same bug or a second one.

## Defects to address (likely related)

1. **(Contained, safe quick-win)** Fix the `get_rx_burst` lookup and logging
   (`src/managers/rdma/daqiri_rdma_mgr.cpp:1777`, client branch `:1787`): use `.find()`
   instead of `rx_rings_map_[conn_id]` so a miss does not insert a null entry (which
   poisons the `:854` gate), and stop emitting `CRITICAL` ~866×/s on the hot path —
   return `NOT_READY` quietly for a legitimately-absent ring and/or demote to `DEBUG`.
   This can land on its own, ahead of the deeper investigation.
2. **(Investigation, may be the same bug)** Explain the steady-state null RX ring and the
   intermittent throughput cliff together. Needs a dedicated trace (per-iteration timing
   in the worker loop, `rx_rings_map_` contents vs the conn_id handed out, completion-queue
   occupancy, credit accounting). The previously-leading "ring-miss predicts the cliff"
   hypothesis is disproven (constant count), but the null-ring lifecycle is the most
   promising starting point.

## Acceptance criteria

- Message-size sweep (`rdma sweep`) shows monotonic-or-better pps as message size
  decreases; no cell collapses to <1 Gb/s without a corresponding drop/error cause.
- Pace sweep (`rdma drop-curve`) tracks every target (incl. 25) within tolerance,
  repeatably across runs.
- `get_rx_burst` no longer emits steady-state `CRITICAL` spam; lookups use `.find()`.

RamyaGuru:
## Summary

Fixes #126: the RoCE/RDMA path intermittently collapsed from ~97 Gb/s to near-zero (~0.001–0.12 Gb/s) on sub-1 MB messages, with **zero CQE errors and zero drops** — so the QP and wire were healthy and the stall was software-side on the op-rate path. The collapse behaved like a race (sub-1 MB almost always, even 8 MB at a specific pace target), which made it hard to pin down.

## Root cause

`rdma_cm` negotiates the QP with **`min_rnr_timer = 0`**, which the IB spec defines as **655.36 ms**, and **`rnr_retry = 7`** (infinite). On fast small messages the sender intermittently outruns the responder's RECV reposting; the responder returns an RNR NAK, and the sender then waits **655 ms** before silently retrying — no CQE error, no drop. Large messages rarely underrun (each occupies the wire long enough for the receiver to repost), which is why the collapse was sub-1 MB and looked intermittent.

Confirmed two ways on the DGX Spark netns wire loopback (Release lib):

- The mlx5 `out_of_buffer` counter rose by only ~47 across a sweep — but at 655 ms each, that's ~31 s of dead time, enough to gut the 30 s small-message cells.
- An `ibv_query_qp` dump showed `timeout=18, retry_cnt=0, rnr_retry=7, min_rnr_timer=0` on both QPs.

## Fix

Lower `min_rnr_timer` to **12 (0.64 ms)** via `ibv_modify_qp` on each QP after the `rdma_cm` connection is established, so a transient RECV underrun costs sub-millisecond instead of 0.65 s. Env-overridable via `DAQIRI_RDMA_MIN_RNR_TIMER`. A one-time INFO log records the resulting QP retry/timeout attributes.

## Results

DGX Spark netns wire loopback, Release, batch 1, 0 drops, 0 CQE errors.

**Message-size sweep**

| message_size | before | after |
| --- | --- | --- |
| 8 MB | 95.9 Gb/s | 101.3 Gb/s |
| 1 MB | 96.1 Gb/s | 100.8 Gb/s |
| 64 KB | 0.118 Gb/s | **10.79 Gb/s** |
| 8 KB | 0.005 Gb/s | **0.935 Gb/s** |
| 4 KB | 0.001 Gb/s | **0.475 Gb/s** |

pps now *rises* as messages shrink (op-rate-bound, not stalled) — the acceptance c

**Pace drop-curve (8 MB):** every target now tracks within ~0.05% with 0 drops, inocked `target=25` (was 0.695 Gb/s in 3/3 runs → now 24.994) and the roaming`target=10` (→ 9.979).

## Also in this PR

- **`get_rx_burst` null-ring fix:** use `.find()` instead of `operator[]` so a miss doesn't insert a null entry that poisons the `rdma_get_server_conn_id` gate; demote the ~866×/s hot-path `CRITICAL` to `DEBUG` and return `NOT_READY`. Verified: spam 25k/cell → 0, TX-coreut regression. (Resolves the contained defect #1 in the issue.)
- **netns server `buf_size` symmetry:** `9000000` → `90000000` to match the client and restore the dropped zero. Flagged by Greptile on #98; confirmed *not* the cliff cause (the cliff was sub-1
MB with 0 drops; a tight buffer bites large messages), justified in the commit mes
- **Perf-doc update:** the published `docs/benchmarks/performance-dgx-spark.md` and the full `docs/performance-dgx-spark.md` now show the fixed RoCE numbers. The published doc carries no
bug/issue framing; the full out-of-nav reference retains the root-cause writeup. Aereqs (`iproute2` + `perftest`) to the Reproduce section.

## Testing

Built Release + `cmake --install`, run in the privileged container as root over thback (PHY-counter verified). Sweep + drop-curve before/after as above.`scripts/check_doc_refs.py` and `mkdocs build --strict` both pass.

## Notes

- Stacked on #98 (`15-bench-spark-perf-doc`) — base this PR on that branch, not `main`.
- Docs-sync: behavioral bugfix only — no public-API/types/config-schema/stream-typDAQIRI_RDMA_MIN_RNR_TIMER`) is an internal tuning env var like the existing`DAQIRI_RDMA_SEND_SIGNAL_EVERY`.

### DIFF
--- a/src/managers/rdma/daqiri_rdma_mgr.cpp
+++ b/src/managers/rdma/daqiri_rdma_mgr.cpp
@@ -498,6 +498,44 @@ void RdmaMgr::rdma_thread(bool is_server, rdma_thread_params* tparams) {
 
   DAQIRI_LOG_INFO("Affined {} RDMA thread to core {}", is_server ? "Server" : "Client", cpu_core);
 
+  // rdma_cm leaves min_rnr_timer at 0, which the IB spec defines as 655.36 ms,
+  // and sets rnr_retry to 7 (infinite). So a single RECV-queue underrun -- which
+  // happens intermittently on fast small messages, where the sender briefly
+  // outruns the responder's RECV reposting -- stalls the sender for 655 ms with
+  // no CQE error and no drop. That is the intermittent small-message throughput
+  // cliff (#126). Lower min_rnr_timer (as responder) so a transient underrun
+  // costs sub-millisecond, not ~0.65 s. Env-overridable for tuning; the default
+  // 12 == 0.64 ms is comfortably above the peer's RECV-repost latency, so it
+  // avoids an RNR-retry storm while keeping any stall negligible.
+  if (tparams->client_id != nullptr && tparams->client_id->qp != nullptr) {
+    int rnr_timer = 12;  // IB min_rnr_timer encoding: 12 == 0.64 ms
+    if (const char* env = std::getenv("DAQIRI_RDMA_MIN_RNR_TIMER")) {
+      const long parsed = strtol(env, nullptr, 10);
+      if (parsed >= 0 && parsed <= 31) { rnr_timer = static_cast<int>(parsed); }
+    }
+    struct ibv_qp_attr mod = {};
+    mod.min_rnr_timer = static_cast<uint8_t>(rnr_timer);
+    if (ibv_modify_qp(tparams->client_id->qp, &mod, IBV_QP_MIN_RNR_TIMER) != 0) {
+      DAQIRI_LOG_WARN("Failed to set min_rnr_timer on {} QP: {}",
+                        is_server ? "server" : "client", strerror(errno));
+    }
+
+    // Confirm the negotiated retry/timeout attributes after the override.
+    struct ibv_qp_attr qattr;
+    struct ibv_qp_init_attr qinit;
+    const int qmask =
+      IBV_QP_TIMEOUT | IBV_QP_RETRY_CNT | IBV_QP_RNR_RETRY | IBV_QP_MIN_RNR_TIMER;
+    if (ibv_query_qp(tparams->client_id->qp, &qattr, qmask, &qinit) == 0) {
+      const double ack_us = qattr.timeout == 0 ? 0.0 : 4.096 * (1U << qattr.timeout);
+      DAQIRI_LOG_INFO(
+        "[rdma-qp {}] timeout={} (local-ack ~{} us; 0=infinite) retry_cnt={} "
+        "rnr_retry={} min_rnr_timer={}",
+        is_server ? "server" : "client", static_cast<int>(qattr.timeout), ack_us,
+        static_cast<int>(qattr.retry_cnt), static_cast<int>(qattr.rnr_retry),
+        static_cast<int>(qattr.min_rnr_timer));
+    }
+  }
+
   // Main TX loop. Wait for send requests from the transmitters to arrive for sending. Also
   // periodically poll the CQ. Exit on either the global force-quit or the per-connection
   // ready_to_exit signal (set by the DISCONNECTED CM handler).
--- a/src/managers/rdma/daqiri_rdma_mgr.cpp
+++ b/src/managers/rdma/daqiri_rdma_mgr.cpp
@@ -510,8 +510,20 @@ void RdmaMgr::rdma_thread(bool is_server, rdma_thread_params* tparams) {
   if (tparams->client_id != nullptr && tparams->client_id->qp != nullptr) {
     int rnr_timer = 12;  // IB min_rnr_timer encoding: 12 == 0.64 ms
     if (const char* env = std::getenv("DAQIRI_RDMA_MIN_RNR_TIMER")) {
-      const long parsed = strtol(env, nullptr, 10);
-      if (parsed >= 0 && parsed <= 31) { rnr_timer = static_cast<int>(parsed); }
+      char* end = nullptr;
+      const long parsed = strtol(env, &end, 10);
+      // Require the whole string to parse as a valid 0-31 code. A bare strtol
+      // would map non-numeric input (e.g. "abc") to 0, which is itself a valid
+      // code meaning 655.36 ms -- silently reintroducing the very stall this
+      // override exists to prevent. Reject and keep the default instead.
+      if (end != env && *end == '\0' && parsed >= 0 && parsed <= 31) {
+        rnr_timer = static_cast<int>(parsed);
+      } else {
+        DAQIRI_LOG_WARN(
+          "Ignoring invalid DAQIRI_RDMA_MIN_RNR_TIMER='{}' (expected integer 0-31); "
+          "using default {}",
+          env, rnr_timer);
+      }
     }
     struct ibv_qp_attr mod = {};
     mod.min_rnr_timer = static_cast<uint8_t>(rnr_timer);
--- a/src/managers/rdma/daqiri_rdma_mgr.cpp
+++ b/src/managers/rdma/daqiri_rdma_mgr.cpp
@@ -1773,27 +1773,26 @@ void RdmaMgr::init_client() {
 }
 
 Status RdmaMgr::get_rx_burst(BurstParams** burst, uintptr_t conn_id, bool server) {
-  if (server) {
-    const auto ring = rx_rings_map_[reinterpret_cast<struct rdma_cm_id*>(conn_id)];
-    if (ring == nullptr) {
-      DAQIRI_LOG_CRITICAL("No server RX ring found for conn_id {:#x}", conn_id);
-      return Status::INVALID_PARAMETER;
-    }
-
-    if (rte_ring_dequeue(ring, reinterpret_cast<void**>(burst)) != 0) { return Status::NOT_READY; }
-
-    return Status::SUCCESS;
-  } else {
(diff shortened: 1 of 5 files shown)


src/managers/rdma/daqiri_rdma_mgr.cpp
```diff
@@ -498,6 +498,56 @@ void RdmaMgr::rdma_thread(bool is_server, rdma_thread_params* tparams) {
...
+  // rdma_cm leaves min_rnr_timer at 0, which the IB spec defines as 655.36 ms,
+  // and sets rnr_retry to 7 (infinite). So a single RECV-queue underrun -- which
+  // happens intermittently on fast small messages, where the sender briefly
+  // outruns the responder's RECV reposting -- stalls the sender for 655 ms with
+  // no CQE error and no drop. That is the intermittent small-message throughput
+  // cliff (#126). Lower min_rnr_timer (as responder) so a transient underrun
+  // costs sub-millisecond, not ~0.65 s. Env-overridable for tuning; the default
+  // 12 == 0.64 ms is comfortably above the peer's RECV-repost latency, so it
+  // avoids an RNR-retry storm while keeping any stall negligible.
+  if (tparams->client_id != nullptr && tparams->client_id->qp != nullptr) {
```
cliffburdick, line 510: In the grand scheme of the benchmark the numbers didn't change much. I wonder if it would be better to instead have the sender pace a bit rather than try to send > 100Gbps like it's probably doing?

cliffburdick (approved):

Issue #135, opened 2026-06-09
Harden DPDK split-burst accessors and RX construction


MikesRND:
## Summary

DAQIRI can crash when a DPDK HDS/split burst reaches user code with missing or incomplete segment metadata. This was observed during live-attach testing where TX traffic was already arriving before the RX application started.

The immediate crash surface is in the public DPDK segment accessors, which can dereference `burst->pkts[seg][idx]` without validating that the segment array and packet slot exist.

## Observed behavior

A receiver can fault in:

- `daqiri::DpdkMgr::get_segment_packet_length(...)`
- `daqiri::DpdkMgr::get_segment_packet_ptr(...)`

This occurred under startup pressure with RX drops reported by DAQIRI/DPDK stats.

## Suspected root cause

1. Segment accessors do not fully validate split-burst metadata before dereferencing packet pointers:
   - `idx < burst->hdr.hdr.num_pkts`
   - `seg < burst->hdr.hdr.num_segs`
   - `burst->pkts[seg] != nullptr`
   - `burst->pkts[seg][idx] != nullptr`
   - equivalent reordered-burst metadata

2. RX split-burst construction can continue after allocation failure for a per-segment packet-pointer array.

3. Split mbuf-chain walking assumes every expected segment exists via `mbuf->next`.

## Requested fix

- Make DPDK segment pointer and length accessors bounds/null safe.
- Prevent RX workers from enqueueing incomplete split bursts.
- Drop malformed split packets or bursts in a controlled way.
- Add bounded diagnostics and/or dropped-packet metrics for incomplete split bursts.

## Reproduction context

Observed during live-attach testing where TX traffic was already arriving before the RX application started. RX drops were visible in DAQIRI/DPDK stats before the crash.

## Environment

Container/environment source: https://github.com/MikesRND/rnd-containers/tree/main `framework-dev` `0.1.2` (`a6d6ae0`)

```text
Platform     Ubuntu 24.04.3 LTS (Noble Numbat) . Linux 6.6.87.2
Compilers    GCC 13.3.0 . CUDA 13.1.80
Build        CMake 3.31.12 . Ninja 1.11.1 . Meson 1.3.2 . ccache 4.9.1
Analysis     clang-format 18.1.3 . clang-tidy 18.1.3 . lcov 2.0
Python       3.12.3
Py Packages  vrtigo 0.2.0 (9f70591) . pytest 9.0.3 . scipy 1.17.1
[...]

PR #137, opened 2026-06-09
#135 - Harden DPDK split burst handling

cliffburdick:
Buffers could be leaked in the HDS use case when bursts were incomplete. Ideally we would add a lot of checks for walking the entire packet list, but this is expensive in the data path. Instead, we add the most common places we could see issues.

Closes #135

PR #140, opened 2026-06-10
#135 - Close split-burst accessor/cleanup gaps from PR #137

MikesRND:
## Summary

Follow-up to #137 for #135. This closes the remaining split-burst safety gaps without changing public APIs.

## Changes

- Make DPDK packet and segment accessors return `nullptr` / `0` for out-of-range or missing burst metadata.
- Add early Raw Ethernet queue segment-count validation before DPDK initialization, with a `validate_config()` backstop.
- Harden DPDK RX cleanup helpers against malformed split-burst metadata.
- Document the four-memory-region limit for Raw Ethernet queues.

## Testing

- Tested on target in framework-dev container.
- Closes issue #135
- Exposes another issue to investigate - still do not have parity with holoscan ANO w.r.t. starting up under heavy pressure.
  - Will continue that investigation separately.


MikesRND:
> AGENTS.md should be updated to reflect the four-memory-region limit introduced and enforced in this PR.

To be clear, this PR did not introduce the four-memory-region limit.  It was pre-existing and just documented here.

src/managers/dpdk/daqiri_dpdk_mgr.cpp
```diff
@@ -1681,6 +1730,14 @@ bool DpdkMgr::set_config_and_initialize(const NetworkConfig& cfg) {
 
   if (!this->initialized_) {
     cfg_ = cfg;
+
+    // Preflight: validate_config() is skipped entirely when initialize() fails, so
+    // catch out-of-range queue segment counts here before DPDK setup runs.
+    if (!validate_queue_segment_counts(cfg_)) {
```
dleshchev, line 1736: **Leaks the `num_init` shutdown refcount on this failure path.** `num_init++` runs unconditionally at the top of `set_config_and_initialize` (line 1729), but this early `return false` doesn't undo it. `shutdown()` only runs the real teardown when `--num_init == 0` (line 4675), and only decrements while `initialized_` is set. So: *bad-MR config → returns false (num_init=1) → user fixes config → init succeeds (num_init=2) → one `shutdown()` → num_init=1* — `cleanup_eal()`/ring frees are skipped and `initialized_` stays stuck true (a later in-process re-init then no-ops). The manager is a persistent singleton (`ManagerFactory::get_active_manager()`, not reset on a failed `daqiri_init`), so the retry reuses the same object and the offset sticks. Fix: don’t leave `num_init` incremented on failure — move `num_init++` below this preflight, or `--num_init` before returning. The pre-existing `initialize()`-failed return at line 1750 has the same flaw and is worth fixing in the same pass.

src/managers/dpdk/daqiri_dpdk_mgr.cpp
```diff
@@ -1681,6 +1730,14 @@ bool DpdkMgr::set_config_and_initialize(const NetworkConfig& cfg) {
...
+      DAQIRI_LOG_CRITICAL("Config validation failed");
```
dleshchev, line 1737: Nit: same generic `"Config validation failed"` string as the post-init `validate_config()` gate at line 1759. The specific reason is already logged inside `validate_queue_segment_counts` (queue name + MR count), so it’s harmless — but a distinct message here (e.g. “segment-count preflight failed”) would make the two failure sites easier to tell apart in logs.

src/managers/dpdk/daqiri_dpdk_mgr.cpp
```diff
@@ -4423,7 +4513,12 @@ void DpdkMgr::free_packet_segment(BurstParams* burst, int seg, int pkt) {
     }
     return;
   }
-  rte_pktmbuf_free_seg(reinterpret_cast<rte_mbuf**>(burst->pkts[seg])[pkt]);
+  const int num_segs = std::clamp(static_cast<int>(burst->hdr.hdr.num_segs), 0, MAX_NUM_SEGS);
+  if (seg < 0 || seg >= num_segs || burst->pkts[seg] == nullptr) { return; }
+  if (pkt < 0 || pkt >= static_cast<int>(burst->hdr.hdr.num_pkts)) { return; }
+  if (burst->pkts[seg][pkt] == nullptr) { return; }
+  rte_pktmbuf_free_seg(reinterpret_cast<rte_mbuf*>(burst->pkts[seg][pkt]));
```
dleshchev, line 4520: Freeing the segment nulls the metadata slot but doesn’t unlink the mbuf from the packet’s `->next` chain. Split-burst segments are chained (seg0->next = seg1…), so a later `free_packet()`/`free_all_packets()` walks the head chain in `dpdk_free_packet_chain` (line 314) straight into this already-freed segment → double-free / UAF. The common cleanup path (`free_all_packets` then `free_rx_burst`) doesn’t mix strategies, and freeing seg 0 first is safe (head becomes null → chain-free takes the orphan fallback), so this only bites a **non-head** per-segment free followed by a whole-packet free. Either detach the segment when freeing it (patch the prior segment’s `->next`), have the chain-free stop at a null slot, or document that per-segment and whole-packet frees are mutually exclusive for a given packet.

dleshchev (commented):
Follow-up review on the split-burst hardening. Core fixes are sound and verified: the segment-count preflight prevents a real OOB write (`rx_nseg = mrs_.size()` feeding the fixed `pkts[MAX_NUM_SEGS]` arrays) and is correctly placed before `initialize()`; the chained-free helper and the accessor bounds-checks are correct.

Three inline notes: (1) the preflight failure path leaks the `num_init` shutdown refcount — the main one to fix; (2) per-segment frees don’t unlink `->next`, so mixing them with whole-packet frees can double-free; (3) a cosmetic duplicate log string.

Confirmed separately: `validate_queue_segment_counts` is DpdkMgr-only and every raw/DPDK example config uses 1–4 memory regions, so no existing config regresses — the 0-MR queues in `daqiri_bench_rdma_*` are `socket`/`roce` (RdmaMgr) and never reach this check.

src/managers/dpdk/daqiri_dpdk_mgr.cpp
```diff
@@ -278,6 +278,55 @@ static bool is_valid_segment_count(int num_segs) {
...
+// any orphaned segment mbufs left behind by a partially-built split burst.
+static void dpdk_free_packet_chain(BurstParams* burst, int pkt, int num_segs) {
+  if (burst->pkts[0] != nullptr && burst->pkts[0][pkt] != nullptr) {
+    rte_pktmbuf_free(reinterpret_cast<rte_mbuf*>(burst->pkts[0][pkt]));
+    for (int seg = 0; seg < num_segs; ++seg) {
+      if (burst->pkts[seg] != nullptr) { burst->pkts[seg][pkt] = nullptr; }
+    }
+    return;
+  }
+  for (int seg = 1; seg < num_segs; ++seg) {
```
cliffburdick, line 322: This shouldn't be necessary. `rte_pktmbuf_free` already frees all segments in a chained mbuf. The only thing left to do here would be to set them to nullptr

PR #149, opened 2026-06-10
Pin example app threads from config

cliffburdick:
Previously only the daqiri queue cores pinned themselves. This opens it up to the app cores as well.

### DIFF
--- a/docs/benchmarks/raw_benchmarking.md
+++ b/docs/benchmarks/raw_benchmarking.md
@@ -55,6 +55,8 @@ docker run --rm -it --privileged \
     - [`daqiri_bench_raw_tx_rx_spark.yaml`](https://github.com/nvidia/daqiri/blob/main/examples/daqiri_bench_raw_tx_rx_spark.yaml) for `daqiri_bench_raw_gpudirect` — still set `eth_dst_addr` to the RX MAC. The rx_port is `0002:01:00.1` (physical port p1), so read its MAC: `cat /sys/class/net/enP2p1s0f1np1/address`. This p0-to-p1 pairing is intentional for an over-the-wire single-machine loopback; using two PFs that map to the same physical port exercises the on-chip eswitch path instead.
     - [`daqiri_bench_rdma_tx_rx_spark.yaml`](https://github.com/nvidia/daqiri/blob/main/examples/daqiri_bench_rdma_tx_rx_spark.yaml) for `daqiri_bench_rdma` — no further edits needed.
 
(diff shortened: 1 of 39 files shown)


RamyaGuru:
Validated on DGX Spark (GB10), Release build, over the physical p0→p1 wire loopback. Ran an A/B of the example bench with app-thread pinning on (this PR's config) vs. off (same config with the `bench_tx`/`bench_rx` `cpu_core` lines stripped), DPDK raw GPUDirect, 0 drops throughout.

**At the native 8000 B shape (wire-bound, ~line rate):** no measurable difference — both ~97 Gb/s with <0.5% run-to-run spread. Expected: the NIC is the ceiling here, so the bench worker thread isn't on the critical path.

**At 256 B (CPU-bound):** the effect is large.

| 256 B | mean throughput | run-to-run spread (5 runs) |
|---|---|---|
| **pinned** (this PR) | 18.1 M pps / 46.4 Gb/s | 18.03–18.17 M pps (~0.8%) |
| **unpinned** | 9.2 M pps / 23.6 Gb/s | 8.08–10.15 M pps (~22%) |

Pinning roughly **doubles** small-packet throughput and **collapses run-to-run variance from ~22% to <1%**.

Why ~2×: the GB10 is heterogeneous — 10 Cortex-X925 (big, 3.9 GHz) + 10 Cortex-A725 (efficiency, 2.8 GHz) cores, and `isolcpus=16-19` reserves four X925 cores. Unpinned, the bench worker is confined to cores 0–15 (the isolated 16–19 are reserved), a mix of X925 and A725 cores that also carry IRQs and OS work. On this heterogeneous CPU the scheduler frequently places/migrates it onto the slower A725 efficiency cores, so it runs at roughly half the big-core rate with high run-to-run variance. Pinning to an isolated X925 core removes the little-core placement, the migration stalls, and the IRQ contention together — hence ~2× and not just the 1.4× clock ratio. The ~22% unpinned spread also matches the "±20% noisy at 64/256 B" we'd previously seen in the Spark perf-doc small-payload cells — this PR is what makes those cells reproducible.

One observation, not a blocker: on a 4-isolated-core box the app worker ends up sharing a core with the DAQIRI queue poller, which costs ~1% at line rate (co-tenancy) but is hugely net-positive in the CPU-bound regime. Worth a follow-up to let the app thread take a dedicated isolated core when one is free.

LGTM from the validation side — good to merge. We'll re-run the DGX Spark perf doc with pinning on; the small-payload DPDK cells will shift up and tighten, while the large/line-rate and RoCE cells are unchanged.

RamyaGuru (approved):

PR #158, opened 2026-06-11
#158 - Style updates to website images and removal of news

awthomp:
Update DAQ and benchmark images to be of the same style, enlargeable, and with the same green border as other images. This also works well with light/dark themes. Removed news section from website; we can add it back in after release.


chloecrozier (approved):
A minor bug fix I put in, was adding the second overlay broke focus on the hero graphic since querySelector was grabbing the overlay’s close button instead of the DAQIRI one

The docs and diagrams look good to me after locally checking!

chloecrozier:
@greptileai review

PR #160, opened 2026-06-12
#129 - reject mixed standard and flex-item RX flows

### LINKED ISSUE #129: [BUG] Mixed standard + flex-item RX flows on one interface are ambiguous
## Summary

When a single RX interface is configured with **both** standard flows and flex-item flows, the DPDK flow setup is ambiguous: group 0 ends up with two all-Ethernet jump rules at the same (default) priority, so only one takes effect per packet and the other flow class — together with its send-to-kernel fallback — is effectively dead.

## Detail

The two flow paths each install their own all-Ethernet jump out of group 0:

- Standard flows: `add_flow()` creates a `group 0 → group 3` jump (`src/managers/dpdk/daqiri_dpdk_mgr.cpp:2773-2774`, pattern = ETH spec=0/mask=0).
- Flex flows: `create_flex_flow_rule()` creates a `group 0 → group 1` jump (`:2568-2571`, same all-Ethernet pattern).

Both jump rules match every packet at the default priority (0). A packet can only jump once, so in a mixed config one jump wins and the other group is never reached — meaning that entire flow class (and the corresponding fallback added in #125) silently does nothing.

PR #125 added symmetric handling:

```cpp
if (intf.rx_.flow_isolation_) {
  if (has_standard_flows && \!add_send_to_kernel_fallback(intf.port_id_, 3)) { return; }
  if (has_flex_item_flows && \!add_send_to_kernel_fallback(intf.port_id_, 1)) { return; }
}
```

This makes it *look* like standard and flex flows are independently supported on the same interface, when in practice mixing them is ambiguous.

## Suggested fix

At minimum, document the limitation (code comment near the flow loop and/or a known-limitation note in the config docs) that a single RX interface should not mix standard and flex-item flows. A more complete fix would detect a mixed config and reject it with a clear error, or restructure the group-0 jump rules so both classes are reachable deterministically (e.g. distinct priorities / match criteria).

## Context

Follow-up from the review of #125 (DPDK flow isolation ordering). The PR mirrors the existing jump structure rather than introducing the ambiguity, but its symmetric `has_standard_flows` / `has_flex_item_flows` handling is what surfaced the concern.

chloecrozier:
Closes #129

On one RX interface, standard (UDP/IP) and flex-item flows install conflicting NIC steering rules, so only one type works and the other fails silently.

**Changes:**
- Fail init with a clear error when both flow types are on the same interface
- Run config validation before DPDK setup
- Document the one-flow-type-per-interface limit

**Test:** start with a YAML that mixes `udp_dst` and `flex_item_id` flows on one interface; init should fail before NIC setup.

**Future work:**
- Support both flow types on one interface by fixing group-0 jump rules (needs mlx5 HWS testing)
- Install the standard-flow group-0 jump once per interface, not once per flow


dleshchev (changes requested):
Review findings after checking the current PR head against origin/main. This does solve #129 by rejecting mixed standard and flex-item RX flows, but the validation move introduces a refcount failure path that should be fixed before merge. Code quality: 4/5. Current examples/benchmarks compatibility: 5/5; behavior-surface impact is low because current examples do not exercise flex-item/standard mixing. Affected applications: raw DPDK/Raw Ethernet consumers using RX flow steering, especially long-lived embedders that retry daqiri_init in-process.

src/engines/dpdk/daqiri_dpdk_engine.cpp
```diff
@@ -1696,6 +1696,12 @@ bool DpdkEngine::set_config_and_initialize(const NetworkConfig& cfg) {
 
   if (!this->initialized_) {
     cfg_ = cfg;
+
+    if (!validate_config()) {
+      DAQIRI_LOG_CRITICAL("Config validation failed");
+      return false;
```
dleshchev, line 1700: This new early return runs after num_init++ but before initialized_ can become true. On validation failure, shutdown() will hit the !initialized_ guard and never decrement num_init, so a later in-process retry can leave EAL/hugepage cleanup skipped on shutdown. Please move the increment until after validation succeeds, or roll it back on every pre-initialized failure path.

chloecrozier:
Thank! Moved num_init++ to after validate_config() succeeds and roll it back on initialize() failure, so validation errors no longer leak the refcount

dleshchev (approved):
Re-reviewed the update in cf7469b. The num_init leak from the pre-init validation failure path is addressed by validating before incrementing, incrementing only once initialization is about to start, rolling back on initialize() failure, and keeping repeat initialization calls counted in the already-initialized branch. No new issues found in the update. The PR still solves #129. Code quality: 4/5. Current examples/benchmarks compatibility: 5/5; behavior-surface impact remains low because current examples do not exercise mixed standard/flex-item RX flows. Verification: git diff --check and scripts/check_doc_refs.py passed.

PR #161, opened 2026-06-12
#128 - Harden DPDK RX flow creation error handling

### LINKED ISSUE #128: [BUG] DPDK RX flow creation failures are silently ignored
## Summary

The DPDK RX flow-creation loop in `DpdkMgr::initialize()` ignores whether each flow was actually programmed into the NIC. A specific flow can silently fail to be created while the send-to-kernel fallback (added in #125) succeeds — the result is that **every packet is quietly delegated to the Linux kernel and the application receives nothing, with no hard error surfaced**.

## Detail

In `src/managers/dpdk/daqiri_dpdk_mgr.cpp` (the loop around `2386-2395`):

```cpp
for (const auto& flow : rx.flows_) {
  DAQIRI_LOG_INFO("Adding RX flow {}", flow.name_);
  if (flow.match_.type_ == FlowMatchType::FLEX_ITEM) {
    add_flex_item_flow(intf.port_id_, flow.match_.flex_item_match_, flow.action_.id_);
    has_flex_item_flows = true;
  } else {
    add_flow(intf.port_id_, flow);
    has_standard_flows = true;
  }
}
```

Neither return value is checked:
- `add_flow()` returns `struct rte_flow*` and yields `nullptr` on failure — the result is discarded.
- `add_flex_item_flow()` only `printf`s on validation/creation failure and otherwise returns the (possibly null) flow.

Because PR #125 now installs a catch-all `SEND_TO_KERNEL` fallback at priority 100 in the same group, a config where the *specific* flow failed to create but the fallback succeeded will look healthy at startup, route all traffic to the kernel, and deliver zero packets to the app.

## Suggested fix

Harden the flow loop to match the new fallback path: check the return of `add_flow()` / `add_flex_item_flow()`, log a CRITICAL on failure, and either abort `initialize()` (`return;`, consistent with the `EalCleanupGuard` + `initialized_` mechanism) or at minimum surface a hard error so the silent kernel-only state is detectable.

## Context

Follow-up from the review of #125 (DPDK flow isolation ordering). Largely pre-existing, but adjacent to the path that PR hardened — worth fixing while the area is fresh.

chloecrozier:
Closes #128
Broken RX flow rules could fail silently while the send-to-kernel fallback still installed, so init looked fine but the app received zero packets.

## Changes

- Abort `initialize()` when `add_flow()` or `add_flex_item_flow()` returns `nullptr`
- Set `has_standard_flows` / `has_flex_item_flows` only after successful flow create
- Add `ensure_eth_jump_rule()` (once per port+group) with `destroy_eth_jump_rules()` on shutdown
- CRITICAL logging for RX, flex, TX offload, and drop-all-traffic flow failures
- `validate_config` rejects flow `action.id` that does not match an RX queue
- `apply_tx_offloads` fails init when `tx_eth_src` flow create fails
- Doc sync: configuration reference, walkthrough, getting-started, concepts, README, AGENTS

## Test

- [x] Container build of `daqiri_dpdk`:

```bash
docker run --rm -v "$(pwd)":/work -w /work daqiri:local \
  cmake --build build -j --target daqiri_dpdk
```
  
## Follow-up (not in this PR)
- Store and rte_flow_destroy() per-flow RX handles and TX offload handles on shutdown
- Replace create_flex_flow_rule() static singleton with per flex_item_id handles
- Reject mixed standard + flex-item RX flows in validate_config()
- Remove unused DropTrafficConfig::jump


chloecrozier:
@greptileai review

dleshchev (commented):
Review findings after checking the current PR head against origin/main. This solves #128: failed RX flow, fallback, and TX-offload programming now abort initialization instead of silently falling through. I found one docs/code ordering issue inline. Code quality: 4/5. Current examples/benchmarks compatibility: 5/5; affected surface includes raw DPDK examples/benchmarks that use rx.flows, flow_isolation, or tx_eth_src, including gpudirect raw, HDS, reorder, quantize, multi-queue, and tx/rx configs. The remaining static flex-item handle issue is real but already tracked separately in #166.

docs/getting-started.md
```diff
@@ -200,6 +200,11 @@ Both methods use the same public C++ include:
 | `DAQIRI_ENABLE_S3` | `OFF` | Enable AWS SDK-backed asynchronous raw packet writes to S3. |
 | `BUILD_SHARED_LIBS` | — | Build as shared library. |
 
+For Raw Ethernet (`stream_type: "raw"`), `daqiri_init()` validates that each `rx.flows`
+entry's `action.id` matches an `rx.queues` ID on the same interface, then programs flow
```
dleshchev, line 204: This says validation happens before programming, but set_config_and_initialize() still calls initialize() first and validate_config() afterward. The new action.id check therefore runs after the NIC flow programming attempt, so a bad queue id may surface as a low-level flow creation failure before this validation message. Either move validation earlier or soften the docs wording to avoid promising validate-then-program order.

chloecrozier:
Thanks, similar to #129! I moved validate_config() before initialize() so the action.id check runs ahead of NIC flow programming. Also rolled back num_init on pre-init failure paths.

dleshchev (approved):
Re-reviewed the update in 9eda93f. Moving validate_config() ahead of initialize() fixes the docs/code ordering issue for action.id validation, and the num_init handling now avoids the pre-init failure leak for validation and initialize() failure paths. No new issues found in the update. The PR still solves #128 by surfacing RX flow, fallback, and TX offload programming failures instead of silently continuing. Code quality: 4/5. Current examples/benchmarks compatibility: 5/5; affected behavior surface remains raw DPDK examples/benchmarks using rx.flows, flow_isolation, or tx_eth_src. Verification: git diff --check and scripts/check_doc_refs.py passed.

Issue #17, opened 2026-04-16
[FEA] Benchmark DAQIRI on RTX Pro Server (C++)


awthomp:
**Describe the solution you'd like**
Backend - DPDK, HDS
- Default yaml config file + FFT
- Default yaml config file + GEMM

Backend - RoCE
- Default yaml config file + FFT
- Default yaml config file + GEMM

Show:
- Total data rate
- Total packets sent
- Total packets received
- Total dropped

PR #162, opened 2026-06-12
#17 - Add RTX PRO 6000 benchmark YAMLs and baseline

chloecrozier:
## What this does

Makes the RTX PRO 6000 benchmarks measure the cable. The branch described a
machine with no loopback, so the wire configs pointed at the wrong port pair and
the baseline held placeholder numbers. There is a 400 GbE loopback cable now, and
every figure below is a closed loop over it, read from the NIC's own PHY counters.

The results are published as
[`docs/benchmarks/performance-rtx-pro-6000.md`](docs/benchmarks/performance-rtx-pro-6000.md),
alongside the existing DGX Spark page. `examples/rtx_pro_6000_baseline.md` keeps
the host-specific record — topology, prerequisites, how to re-run — and defers
every number to that page, so there is no second copy to drift.

## Results

400 GbE loopback, pause off, MTU 9000, median of 3 repetitions (4 for the UDP
ladder). "Cable" is the rate the link sustained, from the sending port's byte
counter; "App" is what the program consumed.

| Benchmark | What it does | Cable Gb/s | App Gb/s |
|---|---|---|---|
| `rdma` (RoCE RC) | 8 MB RDMA writes. The NIC's own engine moves the data — no software sender loop. | **396.9** | 383.5 |
| `dpdk` GPUDirect | 8000 B UDP built in GPU memory, pulled straight from the GPU onto the cable, written straight into another GPU. No CPU touches a payload. | **390.5** | 373.1 |
| `ibverbs` (MPRQ) | The same traffic through the striding-receive-queue ibverbs engine instead of DPDK. | **390.4** | 260.8 |
| `dpdk-hds` | The same journey, but the receiving NIC splits each packet: header to CPU, payload to GPU. | 389.0 | 375.2 |
| `socket-tcp` | Ordinary Linux TCP, 1 MiB writes. | 64.3 | 63.2 |
| `socket-udp`, flat out | Ordinary Linux UDP, 65507 B datagrams, sender unthrottled. | 47.1 | 31.1 |
| `socket-udp`, paced to 25 | The same, sender held to 25 Gb/s. | 25.1 | 25.0 |

Every run crossed the cable with zero frames lost and zero pause frames — now a
gate rather than a column, so a run failing either is a failure, not a slow result.

RoCE holds 99.2% of the port with the *smallest* frames of the fast backends, so
frame size is not what separates these results; gaps in the sender are.
`socket-udp` is the one backend whose sender outruns its own receiver — the port
takes all 46.5 Gb/s off the wire while the application reads 31.1, all of the
difference being datagrams discarded *after* the frames were safely off the wire.

### How much UDP this receiver can take

| Sender held to | Read Gb/s | Clean reps | Worst rep |
|---|---|---|---|
| 20 Gb/s | 20.0 | 3 of 3 | 0 |
| **25 Gb/s** | **25.0** | **4 of 4** | **0** |
| 28 Gb/s | 28.0 | 3 of 4 | 130,371 lost |
| 30 Gb/s | 24.3 | 0 of 4 | 193,792 lost |
| flat out | 35.4 | 0 of 4 | 364,807 lost |

**25 Gb/s is the number to quote.** 28 is where it stops being dependable: three
repetitions lost nothing and the fourth lost 16% of its datagrams. The median for
that rung is a flat zero and hid it, which is why the ladder counts clean
repetitions instead.

## Engine change

`socket_config` gains `rx_buffer_size` / `tx_buffer_size`, applied at socket
creation and exposed to Python. Without them a UDP receiver gets
`net.core.rmem_default` — 256 KB, under four of the 65507 B datagrams this test
sends — and discards the overflow where no NIC counter and no sender sees it. A
kernel clamp is logged, since being silently capped is how a buffer that looks
configured still drops.

## Measurement fixes behind these numbers

Four ways a run could report something other than what it measured, all fixed
here: `LD_LIBRARY_PATH` searched `/opt/daqiri/lib` before the build tree, so runs
measured the container image's copy (which is why `ibverbs` RX looked broken — the
fix was in the source, never the binary); the namespace path dropped
`--target-gbps`, so every drop-curve rung ran flat out under a different label; a
cell that never reached its target drops nothing and read as the cleanest point on
the curve, so the NIC's transmit rate is now checked against the target; and
dividing the `ibverbs` cell's bytes by its shorter-lived receiver's window gave
472 Gb/s on a 400 GbE port, so rates are now checked against line rate.
"Sustained" was also the highest single one-second sample; it is now a mean.

## Known gaps

- `ibverbs` RX consumes 260.8 Gb/s against its own transmit side's 390.4.
- The UDP receive path allocates and copies per datagram (1.26 us + 0.256 ns/byte),
  holding it near 31 Gb/s. Receiving into pooled buffers instead lifts it to
  42–46 Gb/s, but that is an engine change rather than a benchmark one and is kept
  on a follow-up branch so this PR stays scoped to the harness.
- No `isolcpus` here, so spread is a few percent on the fast backends, wider on
  `socket-udp`.
- 800 Gb/s is only a bidirectional aggregate; nothing in the harness drives both
  ways.


docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+`carrier=1` on two ports does not make them a pair — this host has a second,
+uncabled ConnectX-7 whose ports also show carrier.
+[`scripts/discover_rtx_pro_topology.sh`](https://github.com/nvidia/daqiri/blob/main/scripts/discover_rtx_pro_topology.sh)
+identifies the cabled pair from mutual LLDP neighbours (or, while the namespaces
+are up, from the namespaces themselves) and fills the PCIe addresses, destination
+MAC, PIX GPU ordinals, and poll cores into each config.
+
+## Results summary
+
+Median of 3 repetitions per cell (4 for the UDP ladder), measured 2026-08-25.
```
cliffburdick: Do we need to include a date here?

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+
+| Stream / Protocol | Best case | Sustained on the cable | App consumed | Lost on the cable |
+| ----------------- | --------- | ---------------------: | -----------: | ----------------- |
+| Socket / RoCE (SEND) | 8 MB message | **396.9 Gb/s** (99.2% of the port) | 383.5 | 0 |
```
cliffburdick: In the future we should probably include the same size as raw ethernet, otherwise these comparisons are a bit of apples to oranges. RoCE will have the advantage here just due to the message size chosen.

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+Every run above carried every frame it sent and saw no pause frames — both are
+enforced as gates, so a run that loses frames on the cable or gets flow-controlled
+is recorded as a failure rather than as a slow result. The unpaced UDP row is the
+only one where the two rate columns are far apart: its sender put 46.3 Gb/s of
+payload on the wire and the receiving application read 31.1, all of the difference
+discarded inside the host *after* the frames were safely off the cable. That gap
+is the subject of [Socket / UDP](#socket-udp) below.
+
+**Frame size is not what separates these results.** RoCE holds the highest rate
+with the *smallest* frames of the fast backends (4057 B on the wire, against
```
cliffburdick: This is true, but as mentioned above RoCE also has an 8MB message submitted to it rather than a bunch of 4096B packets, which gives it an advantage. If RoCE were instead given a message size of 8K the performance will drop due to its overhead.

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+## How the numbers are measured
+
+Each run snapshots `ethtool -S` on both ports before and after, and reports from
+the `*_phy` counter family. Those are the hardware's own tallies, so they do not
+depend on the benchmark's bookkeeping or on when it started its timer.
+
+| Figure | What it is |
+| ------ | ---------- |
+| **Sustained on the cable** | The sending port's byte counter sampled once a second and averaged over the run, discarding the first and last (partial) samples. A mean, not a peak: a maximum is one sample and says nothing about what the link held. It counts the frame, CRC, preamble and interframe gap — everything that occupies the link — so it is directly comparable to the port's 400 Gb/s. |
+| **App consumed** | What the receiving program itself reported. Shown so the two can be compared: where they disagree, the difference is being lost inside the host, after the wire already delivered it. |
```
cliffburdick: This is otherwise known as "goodput" and can be mentioned here. The "Sustained on the cable" is "throughput".

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+| **Lost on the cable** | Frames the sending port emitted minus frames the receiving port took off. Any non-zero value implicates the link itself. |
```
cliffburdick: Maybe just "Lost packets" rather than "Lost on the cable"?

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+| **Pause frames** | 802.3x flow control seen by either port. Pause throttles a sender without incrementing any drop counter, so a paused run looks clean and is simply slow. One pause frame fails the run. |
+
+A **paced** run gets one more check. If the NIC's transmit rate lands more than 5%
+under the rate the sender was told to hold, the cell is discarded rather than
```
cliffburdick: or over?

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+receiving port writes them straight into GPU 2's memory. No CPU reads or copies a
+payload.
+
+The cable sustains **390.5 Gb/s**, 97.6% of the port. Over 20 s the sending port
+put 115,783,681 frames on the wire and the receiving port took exactly
+115,783,681 off, and the three repetitions held 390.5, 390.4 and 390.5 Gb/s with
+the two ports agreeing to within 0.01 Gb/s in each. The receiver reports ~5,500
+dropped packets out of 115 million (0.005%), all of it host-side.
+
+### Payload sweep (best batch per payload, 15 s per cell)
```
cliffburdick: I would add in the title that these are single-core results

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+Then the cells themselves. Each writes `bench-results/<timestamp>-<backend>-<mode>/`
+with a `runs.csv` carrying `wire_tx_gbps` / `wire_rx_gbps` per cell, and a
+`wire_validation.txt` holding the raw NIC frame and byte counts, what was lost,
+and any pause frames:
+
+```bash
+# Namespaces down:
+./examples/run_rtx_pro_bench.sh dpdk     nic-smoke --seconds 30
+./examples/run_rtx_pro_bench.sh dpdk     sweep     --seconds 15
+./examples/run_rtx_pro_bench.sh dpdk     issue17   --seconds 20
```
RamyaGuru: could we rename this flag from "issue17" to something more descriptive?

examples/rtx_pro_6000_baseline.md
RamyaGuru, line 1: I think this file doesn't need to published, but could be helpful for internal use. So maybe we remove it form the PR. Looks like it has more so the implementation details for the internal cluster. Correct me if I'm wrong, @chloecrozier

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+| 8000 B | 372.3 | 371.9 | 1024 |
+| 4096 B | 372.0 | 371.5 | 1024 |
+| 1024 B | 355.5 | 259.2 | 10240 |
+| 256 B  | 100.6 | 77.8  | 1024 |
+| 64 B   | 33.7  | 31.3  | 10240 |
+
+Down to 4096 B the two columns agree and the link is the limit. Below that they
+part company — and the cable figure falls too, because with one queue per
+direction small packets bottleneck the *sender* as well. That is why 64 B reaches
+only 33.7 Gb/s on the wire here, while two send queues push 111.8 Gb/s at the same
```
cliffburdick: In future benchmarks we should separate out sender from receiver so they can assume the transmitter is always matching what the receiver can do.

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+size. At 1024 B the gap is purely on the receive side: the cable carries 355.5 and
+one receive core takes 259.2.
+
+### Multi-queue core scaling (20 s per cell)
+
+Each cell is **cable Gb/s / app Gb/s** — what the NIC carried, and what the
+receiving program read out of it. `2t1r` is two send queues and one receive queue,
```
cliffburdick: Does this mean one core per queue or is it one core across all queues?

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+| **Lost on the cable** | Frames the sending port emitted minus frames the receiving port took off. Any non-zero value implicates the link itself. |
+| **Pause frames** | 802.3x flow control seen by either port. Pause throttles a sender without incrementing any drop counter, so a paused run looks clean and is simply slow. One pause frame fails the run. |
+
+A **paced** run gets one more check. If the NIC's transmit rate lands more than 5%
+under the rate the sender was told to hold, the cell is discarded rather than
+averaged in. Without it, a cell that lost its transmit core and only managed
+15 Gb/s still reports zero loss under a "paced to 20" label — because the receiver
+was never pushed — and reads as the cleanest point on the curve.
+
+## Raw Ethernet / GPUDirect (DPDK)
```
cliffburdick, line 160: DPDK is deprecated at this point with ibverbs being the default backend. We can include it, but ibverbs should be listed first if possible.

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+receive side: at 1024 B with two send queues the cable carried 372.2 while one
+receive core took 255.7, and a second receive core recovers nearly all of it
+(372.2). A second *send* core alone only widens the gap, since it pushes harder at
+a receiver that already could not keep up.
+
+So a low number at a small payload is not the link running out. It is one core
+running out, and the frames it could not take are dropped inside the host after
+the wire had already delivered them.
+
+### GPU work on the received data (8000 B, 3 × 20 s)
```
cliffburdick, line 231: The payload should probably be 8192B here. In the FFT and GEMM cases the non-power-of-two will result in poorer performance when it might be able to keep up.

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+
+## Raw Ethernet / header-data split
+
+The same journey, except the receiving NIC cuts each arriving packet in two — the
+header to CPU memory, the 8000-byte payload to GPU memory. This is what you want
+when software has to read sequence numbers or timestamps while the payload never
+leaves the GPU. It costs essentially nothing: **389.0 Gb/s** on the cable against
+GPUDirect's 390.5, with 375.2 consumed.
+
+## Raw Ethernet / ibverbs (MPRQ)
```
cliffburdick: You can remove MPRQ since DPDK has this as well

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+
+The same traffic through the [ibverbs multi-packet receive queue](raw_benchmarking.md)
+instead of DPDK, as a DPDK sender feeding an MPRQ receiver over the cable.
+
+Sending is not the problem: the transmit port holds **390.4 Gb/s**, matching
+DPDK. The MPRQ receive side consumes **260.8**, dropping 65.5 M of 146 M packets.
```
cliffburdick: If this is saying ibverbs is lower than DPDK we probably should investigate this before publishing since it's the opposite of what I was measuring. Is this with device memory?

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+    Every other backend here is one process whose own timer bounds the traffic, so
+    dividing the NIC's byte totals by it gives the rate. This cell is two
+    processes, and the sender both starts earlier and outlives the receiver's
+    measured window — dividing all the bytes by only the receiver's 20 s produced
+    472 Gb/s on a 400 Gb/s port. The runner now checks every rate against line
+    rate and falls back to the full snapshot window, which spans start-up and
+    teardown and therefore understates. The sampled sustained figure is unaffected
+    and is the one to use.
+
+## Socket / RoCE
```
cliffburdick, line 264: Maybe say RoCE SEND/RECEIVE here since the terminology is for that

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+Two processes over the namespace wire loopback, one queue pair. The client posts
+8 MB RC SENDs; the NIC splits each into 4 KB frames (the RoCE path MTU under a
+9000-byte netdev MTU), and the receiving HCA places the bytes directly into the
+server's pre-posted receive memory. Nothing runs per-packet on either CPU, which
+is why this is the fastest backend on the page: **396.9 Gb/s sustained**, 99.2% of
+the port, with 383.5 consumed.
+
+## Socket / TCP
+
+Ordinary Linux TCP, written in 1 MiB chunks, one kernel thread per side doing the
```
cliffburdick: Are we sure a single thread achieved this? We had 30Gbps on the Spark slide

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+the kernel stack is simply not a 400 GbE transport; the comparison worth drawing
+is against RoCE on the same cable, six times faster with smaller frames.
+
+## Socket / UDP
+
+Ordinary Linux UDP: the client sends 65507-byte datagrams, the kernel splits each
+into 8 IP fragments to fit the 9000-byte MTU, and the far side reassembles. UDP
+never retransmits, so whatever the receiver cannot keep up with is gone.
+
+This is the one backend whose sender comfortably outruns its own receiver. Flat
```
cliffburdick, line 305: We should eventually investigate this. UDP should outperform TCP easily.

docs/benchmarks/performance-rtx-pro-6000.md
```diff
@@ -0,0 +1,379 @@
...
+  unpaced sender leaves a deep queue and amortises the syscall across a full
+  batch, while a smoothly paced sender delivers closer to one at a time and pays
+  the per-call cost on each. Across all seven unpaced repetitions the read rate
+  came out at 25.6, 31.1, 32.7, 33.5, 34.7, 36.1 and 45.9 Gb/s. That spread is
+  wider than the gap between two rungs of the ladder, which is why the
+  recommendation is 25 rather than the highest number that ever worked.
+
+## Known limitations
+
+- **`ibverbs` RX consumes 260.8 Gb/s** while its own transmit side holds 390.4.
```
cliffburdick: Let's resolve this before merging since this is not what I was getting

docs/tutorials/configuration-walkthrough.md
```diff
@@ -31,8 +31,13 @@ For a shorter selection guide, start with the [Benchmarking overview](../benchma
...
+    - **RTX PRO 6000 Blackwell wire loopback** (template, replace `<placeholders>`): [`daqiri_bench_raw_tx_rx_rtx_pro_6000.yaml`](https://github.com/nvidia/daqiri/blob/main/examples/daqiri_bench_raw_tx_rx_rtx_pro_6000.yaml). Same shape with PCIe addresses, MAC, and GPU ordinals left blank, for a different RTX PRO host or a cross-card topology.
+    - **RTX PRO 6000 Blackwell software loopback**: [`daqiri_bench_raw_sw_loopback_rtx_pro_6000.yaml`](https://github.com/nvidia/daqiri/blob/main/examples/daqiri_bench_raw_sw_loopback_rtx_pro_6000.yaml). `loopback: "sw"` with `kind: device` buffers. Verifies the GPUDirect build on an RTX PRO host without a cable. It never touches the NIC, so its throughput is not a wire rate and can exceed line rate.
```
cliffburdick, line 37: We should drop software loopback in most cases since it should be deprecated/unused in almost all cases.

docs/tutorials/configuration-walkthrough.md
```diff
@@ -49,7 +54,7 @@ For a shorter selection guide, start with the [Benchmarking overview](../benchma
...
+    - **Netns wire loopback** (prefilled, combined base): [`daqiri_bench_rdma_tx_rx_spark_netns.yaml`](https://github.com/nvidia/daqiri/blob/main/examples/daqiri_bench_rdma_tx_rx_spark_netns.yaml). Carries both roles; `scripts/gen_spark_netns_config.py` splits it per role so each runs in its own network namespace (`--mode server` / `--mode client`) and RDMA-CM resolves over the wire instead of short-circuiting in the kernel. Used by `examples/run_spark_bench.sh` on a Spark and by `examples/run_rtx_pro_bench.sh rdma` on an RTX PRO 6000 — the file is not Spark-specific, and neither is the namespace setup once `CLIENT_IF`/`SERVER_IF` are set. See [Socket and RDMA Benchmarking](../benchmarks/socket_benchmarking.md#run-the-rdma-roce-benchmark); RTX PRO 6000 rates at [Performance: RTX PRO 6000](../benchmarks/performance-rtx-pro-6000.md#socket-roce).
```
cliffburdick, line 57: I think this is also what's needed for socket mode (non-roce), otherwise you're just getting loopback results

examples/rtx_pro_nic_counters.sh
cliffburdick, line 1: This seems generic for all nics. What's the reason it's under an rtx pro name?

examples/rtx_pro_yaml_rewrite.sh
cliffburdick, line 1: Same comment as above

chloecrozier:
Addressed everything except the comments asking for refreshed numbers. Some of those I had run before but never validated, so rather than publish them I left the existing values in place and flagged the tables that still need re-measuring — including the `ibverbs` RX result, so that one is still open.

chloecrozier:
@RamyaGuru @cliffburdick I addressed most of the comments except the ones about the most up-to-date numbers. Some of those I had run in bulk before but never fully verified, so I left current values in there so we can just put the real good numbers when we find them and we can coordinate over slack!

scripts/setup_spark_wire_loopback_netns.sh
cliffburdick, line 1: Same comment as before, where this should probably drop the name "spark"

Issue #177, opened 2026-06-12
[FEA] Remove DPDK as core dependency


cliffburdick:
DPDK should only be needed if the DPDK engine is chosen. Remove any dependency on its rings and pools for other engines.

PR #178, opened 2026-06-12
#177: Drop DPDK as a dependency unless the dpdk engine is built

cliffburdick:
daqiri_common and the rdma/ibverbs/socket engines previously required libdpdk solely for the generic rte_ring (pointer ring) and rte_mempool (fixed-size object slab) primitives, and called rte_eal_init only to back that memory. Replace those with libdpdk-free header-only primitives so a build without the dpdk engine links no librte_*.

- Add daqiri::Ring (src/daqiri_ring.h): a bounded MPMC/SPSC pointer ring, a faithful reimplementation of DPDK's C11 4-cursor rte_ring with all-or-nothing bulk and capacity-1 usable semantics. The acquire fence in the head-reservation loop is load-bearing on weakly-ordered CPUs (aarch64).
- Add daqiri::ObjectPool (src/daqiri_pool.h): a slab + free-Ring replacing the generic rte_mempool (get/put/avail_count/in_use_count).
- Move the DPDK-only Engine base-class methods (mbuf/extmem registration, mbuf packet pools, EAL/hugepage preflight + cleanup) out of engine.cpp into src/engine_dpdk.cpp, compiled into daqiri_common only when dpdk is selected. MemoryKind::HUGE allocation now goes through a virtual alloc_huge hook (mmap(MAP_HUGETLB) base; rte_malloc_socket in DpdkEngine).
- Convert the rdma and ibverbs engines to the new primitives and delete their rte_eal_init/cleanup entirely.
- CMake: parse DAQIRI_ENGINE first, derive DAQIRI_BUILD_DPDK, and only then look up/link libdpdk; drop DPDK from the rdma/ibverbs/socket engine builds.
- Add examples/pool_ring_bench.cpp comparing the primitives against DPDK (rte arm compiled only in DPDK builds); measured parity is 81-111%.

Verified: builds with DAQIRI_ENGINE="ibverbs", "dpdk ibverbs", and "" all pass; readelf confirms no librte_* in the non-DPDK builds; the DPDK sw-loopback bench still runs without regression.


dleshchev (commented):
Reviewed against `origin/main` in an isolated worktree (read-only). Solid decoupling work — the `daqiri::Ring` / `ObjectPool` primitives are carefully implemented (correct C11 4-cursor algorithm, the acquire fence on the aarch64 path is right), the engine conversions look clean, and removing the `#if DAQIRI_ENGINE_DPDK || DAQIRI_ENGINE_RDMA` guard on `allocate_memory_regions()` (`engine.cpp`) incidentally fixes a latent bug where an **ibverbs-only** build allocated no memory regions. Two concrete issues plus smaller notes below.

**1. (blocking) The no-DPDK example build is broken** — see the inline comment on `examples/pool_ring_bench.cpp:367`, with the exact compile command and error. This blocks the `DAQIRI_ENGINE=""` configuration the PR exists to enable when `DAQIRI_BUILD_EXAMPLES=ON`.

**2. Installed `daqiri.pc` still hard-requires libdpdk in non-DPDK builds.** This one isn't in the diff (the top-level `CMakeLists.txt` isn't touched by this PR), so I can't anchor it inline — but it's directly in scope for the stated goal. `CMakeLists.txt:96` sets `DAQIRI_PC_REQUIRES_PRIVATE "libdpdk"` unconditionally, and that flows into the installed `${libdir}/pkgconfig/daqiri.pc` via `cmake/daqiri.pc.in:10`. So a `DAQIRI_ENGINE="ibverbs"`/`""` build still ships `Requires.private: libdpdk`, and consumers then need `libdpdk.pc` present for `pkg-config --cflags daqiri` (Requires.private contributes cflags unconditionally) and `--libs --static`. The library no longer links `librte_*`, but the package metadata still declares the dependency. The neighboring block already gates `libibverbs librdmacm` on the engine list — suggest appending `libdpdk` the same way, only when `dpdk` is in `DAQIRI_ENGINE`.

**Smaller notes:**
- **Doc-sync:** this touches `src/CMakeLists.txt` (new optional `DAQIRI_HAVE_NUMA`, DPDK now conditional) and adds a bench. Per the repo's docs-sync rules that should reach `docs/getting-started.md` (still lists DPDK as a general dependency, no libnuma mention) and `README.md` (Quick Start / Engines) — only `AGENTS.md` was updated. CI's `check_doc_refs.py` won't flag it (the new bench has no YAML).
- **Commit subjects** don't follow the `#177 - <Title>` convention from CONTRIBUTING.md (the PR title has the prefix; the individual commits don't).

examples/pool_ring_bench.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+  struct RingCase {
+    const char* name;
+    int p, c;
+    unsigned bulk;
+    daqiri::RingMode mode;
+    unsigned rte_flags;
+  };
+  const RingCase ring_cases[] = {
+      {"SPSC single 1p1c", 1, 1, 1, daqiri::RingMode::SPSC, RING_F_SP_ENQ | RING_F_SC_DEQ},
```
dleshchev, line 373: **Breaks the no-DPDK examples build.** The `ring_cases[]` initializer references `RING_F_SP_ENQ | RING_F_SC_DEQ` unconditionally, but those macros come only from `<rte_ring.h>`, which is included solely under `#if BENCH_HAVE_DPDK` (lines 46-50). The `#else` branch already treats the field as unused (`(void)rc.rte_flags;`, line 381), but the initializer *values* are still evaluated in every build. Since this bench is built whenever the `daqiri_dpdk` target exists and `BENCH_HAVE_DPDK` is set from a *separate* `pkg_check_modules(... QUIET libdpdk)` probe in the examples scope, it fails to compile in any config where that probe misses DPDK — including the `DAQIRI_ENGINE=""` build this PR exists to enable.

Reproduced with g++ 11.4.0, no `-DBENCH_HAVE_DPDK`:

```
$ g++ -std=c++17 -I <repo_root> -c examples/pool_ring_bench.cpp -o /dev/null
examples/pool_ring_bench.cpp: In function 'int main(int, char**)':
examples/pool_ring_bench.cpp:367:61: error: 'RING_F_SP_ENQ' was not declared in this scope
  367 |       {"SPSC single 1p1c", 1, 1, 1, daqiri::RingMode::SPSC, RING_F_SP_ENQ | RING_F_SC_DEQ},
      |                                                             ^~~~~~~~~~~~~
examples/pool_ring_bench.cpp:367:77: error: 'RING_F_SC_DEQ' was not declared in this scope
  367 |       {"SPSC single 1p1c", 1, 1, 1, daqiri::RingMode::SPSC, RING_F_SP_ENQ | RING_F_SC_DEQ},
      |                                                                             ^~~~~~~~~~~~~
```

(The `daqiri::Ring` / `ObjectPool` headers compiled cleanly — the only errors are the DPDK macros, which confirms the primitives themselves are standalone.)

Fix: gate the flag values, e.g. wrap the SPSC rows in `#if BENCH_HAVE_DPDK … #else … 0 … #endif`, or add `#define RING_F_SP_ENQ 0` / `RING_F_SC_DEQ 0` fallbacks under the existing `#else`.

src/engine.h
```diff
@@ -114,6 +122,10 @@ class Engine {
                                  const std::string& local_mr_name);
   virtual RDMAOpCode rdma_get_opcode(BurstParams* burst);
   int numa_from_mem(const MemoryRegionConfig& mr) const;
+  // mbuf/extmem-based helpers used only by the DPDK engine. Their definitions
+  // live in engine_dpdk.cpp, compiled into daqiri_common only when the dpdk
+  // engine is selected; in non-DPDK builds they are declared-but-undefined and
+  // never referenced (they are only called from DpdkEngine), so they link fine.
   Status register_memory_regions();
```
dleshchev, line 129: Minor / maintainability: these methods (`register_memory_regions`, `map_memory_regions`, `create_{pktmbuf,generic}_pool`, `estimate_required_hugepages`, `check_hugepage_availability`, `cleanup_eal`) are declared here but now defined only in `engine_dpdk.cpp`, which is compiled in DPDK builds only. They're currently called solely from `daqiri_dpdk_engine.cpp`, so it links — but a future non-DPDK caller would surface as a link error rather than a compile error. Worth a short comment marking these as dpdk-only so the contract is explicit.

src/engines/rdma/daqiri_rdma_engine.h
```diff
@@ -193,16 +194,16 @@ class RdmaEngine : public Engine {
...
-  std::queue<struct rte_ring*> tx_rings_;
-  std::queue<struct rte_ring*> rx_rings_;
-  std::unordered_map<struct rdma_cm_id*, struct rte_ring*> tx_rings_map_;
-  std::unordered_map<struct rdma_cm_id*, struct rte_ring*> rx_rings_map_;
-  std::unordered_map<std::string, struct rte_ring*> mem_pools_;
-  struct rte_mempool* tx_burst_pool_;
+  daqiri::ObjectPool* rx_meta;
+  daqiri::ObjectPool* tx_meta;
+  daqiri::ObjectPool* pkt_len_pool_;
+  std::unordered_map<std::string, daqiri::ObjectPool*> mr_pools_;
```
dleshchev: Nit: `mr_pools_` was retyped to `daqiri::ObjectPool*` but appears unused after the conversion — I see no references to it anywhere in `daqiri_rdma_engine.cpp` (0 uses). If it's genuinely dead now, drop it to avoid implying it's live state.

Issue #176, opened 2026-06-12
[FEA] Add packet pacing


cliffburdick:
Add packet pacing to raw backends

PR #179, opened 2026-06-13
#176 Packet pacing

cliffburdick:
Added packet pacing for raw streams. This allows a user to meter packets out at a specific rate within 0.5% of the target rate.


examples/daqiri_bench_raw_tx_rx_pacing.yaml
```diff
@@ -0,0 +1,93 @@
+%YAML 1.2
```
dleshchev, line 1: **This config is never copied or installed.** It isn't listed in `DAQIRI_BENCH_CONFIGS` in `examples/CMakeLists.txt` (the list starting around line 25), and that list drives both the `configure_file(... COPYONLY)` loop (→ `build/examples/`) and `install(FILES ${DAQIRI_BENCH_CONFIGS} ...)` (→ `/opt/daqiri/...`). But `docs/benchmarks/raw_benchmarking.md:224` tells users to run:

```
/opt/daqiri/bin/daqiri_bench_raw_gpudirect /opt/daqiri/bin/daqiri_bench_raw_tx_rx_pacing.yaml --seconds 10
```

So the documented invocation is file-not-found out of the box, in both the from-source (`./build/examples/`) and container (`/opt/daqiri/bin/`) layouts.

Note CI won't catch this: `check_doc_refs.py` only enforces that every YAML is *referenced in the walkthrough tree*, not that it's *installed*.

Fix: add `daqiri_bench_raw_tx_rx_pacing.yaml` to `DAQIRI_BENCH_CONFIGS`.

src/engines/dpdk/daqiri_dpdk_engine.h
```diff
@@ -162,6 +162,11 @@ class DpdkEngine : public Engine {
   bool setup_rx_timestamp_dynfield();
   bool setup_tx_timestamp_dynfield();
   bool calibrate_rx_timestamp_clock(uint16_t port_id);
+  // Current NIC clock for `port` in nanoseconds, for packet-pacing scheduling.
+  // Uses rte_eth_read_clock + the calibrated ticks/sec where supported; falls
+  // back to CLOCK_REALTIME (e.g. mlx5, whose real-time clock tracks UTC and
```
dleshchev: This comment contradicts the implementation. It says `now_tx_ns` "falls back to **CLOCK_REALTIME** (e.g. mlx5, whose real-time clock tracks **UTC** …)", but the implementation does the opposite — `daqiri_dpdk_engine.cpp:1944-1948` states "tracking **CLOCK_MONOTONIC**, **NOT** CLOCK_REALTIME (UTC)" and calls `clock_gettime(CLOCK_MONOTONIC, &ts)`. This looks like a leftover from before the clock-seeding fix. Please update it to CLOCK_MONOTONIC so it doesn't mislead about which host clock the pacing schedule is anchored to.

src/engines/dpdk/daqiri_dpdk_engine.cpp
```diff
@@ -4237,6 +4331,41 @@ int DpdkEngine::tx_core_worker(void* arg) {
...
+      }
+      auto* m0 = reinterpret_cast<struct rte_mbuf*>(msg->pkts[0][0]);
+      m0->ol_flags |= tparams->tx_ts_dynflag_mask;
+      *RTE_MBUF_DYNFIELD(m0, tparams->tx_ts_dynfield_offset, uint64_t*) = tparams->pace_next_ns;
+      // ns = bytes * 8 bits / (Mbps * 1e6 bits/s) * 1e9 ns/s = 8000 * bytes / Mbps.
+      // Carry the division remainder so the long-run average is exactly pacing_mbps.
+      // The set_*_packet_lengths helpers maintain hdr.nbytes as the burst's L2 byte
+      // total; use it directly to avoid walking the burst. Fall back to a walk for
+      // bursts populated without those helpers (nbytes left at 0).
+      uint64_t bytes = msg->hdr.hdr.nbytes;
```
dleshchev, line 4358: The `nbytes == 0` sentinel can't distinguish "helpers weren't used" from a stale recycled value. `hdr.nbytes` is reset to 0 in `set_header` and accumulated in `set_packet_lengths` (`+= ttl_len`), but TX metadata buffers come from a recycled mempool and `nbytes` isn't cleared on recycle — so the running count is correct only if the app always calls `set_header` before `set_packet_lengths`. If lengths are filled on a recycled buffer holding a stale nonzero `nbytes` without calling `set_header` first, pacing over-counts, and the `== 0` check can't detect it. All in-tree benches call `set_header` first so this isn't a live bug, but consider documenting the ordering contract or using an explicit `valid` flag rather than overloading 0.

src/engines/dpdk/daqiri_dpdk_engine.cpp
```diff
@@ -4558,6 +4689,8 @@ Status DpdkEngine::set_all_packet_lengths(BurstParams* burst,
 
   auto** first_seg = reinterpret_cast<rte_mbuf**>(burst->pkts[0]);
   for (int idx = 0; idx < num_pkts; ++idx) { first_seg[idx]->pkt_len = ttl_len; }
+  // Record the burst's L2 byte total so TX pacing need not walk the burst.
+  burst->hdr.hdr.nbytes = ttl_len * static_cast<uint32_t>(num_pkts);
```
dleshchev: Nit: this multiply is done in uint32 (`nbytes` is `uint32_t`), so a very large batch could overflow before the pacing math widens it to uint64. Fine for current batch sizes (~10240 × ~8 KB ≈ 82 MB), but consider `static_cast<uint64_t>(ttl_len) * num_pkts` (or widening the field) to be safe.

dleshchev (commented):
Reviewed against `origin/main` in an isolated worktree (read-only). The DPDK pacing implementation is coherent, and the accuracy-critical piece — seeding `pace_next_ns` from `rte_eth_read_clock()` (the same clock the mlx5 scheduler compares against) rather than a host clock — is the right fix. Removing the non-functional ibverbs WAIT-WQE path rather than shipping one that silently stalls is also the right call. One blocking packaging issue and a few smaller notes are inline.

**Scope vs. issue #176:** the issue says "raw backends" (plural). In this codebase `ibverbs` is *also* a raw-Ethernet (MPRQ) backend, and this PR explicitly rejects `pacing_mbps > 0` there — so coverage is partial against the issue text. That's a reasonable call given the missing send-scheduling infrastructure, but worth stating explicitly in the PR/issue so the narrowed scope is on record.

Otherwise the doc-sync is thorough (config reference, raw_benchmarking, walkthrough decision-tree leaf, AGENTS.md) and the DCO / commit-title conventions are met.

examples/daqiri_bench_raw_tx_rx_pacing.yaml
cliffburdick: Fixed in fdb8500 — added `daqiri_bench_raw_tx_rx_pacing.yaml` to `DAQIRI_BENCH_CONFIGS` in `examples/CMakeLists.txt`, so it is now copied to `build/examples/` and installed to `/opt/daqiri/bin/`. Verified the file appears in `build/examples/` after a build. Thanks for catching the file-not-found in the documented invocation.

src/engines/dpdk/daqiri_dpdk_engine.h
cliffburdick: Fixed in fdb8500 — updated the comment to say it falls back to `CLOCK_MONOTONIC` (matching the implementation), with a note that the NIC PTP clock free-runs from driver load unless a PTP daemon disciplines it, so it tracks MONOTONIC rather than REALTIME/UTC.

src/engines/dpdk/daqiri_dpdk_engine.cpp
cliffburdick: Addressed in fdb8500 — `hdr.nbytes` is now cleared when a TX metadata buffer is acquired (`create_tx_burst_params` and `get_tx_metadata_buffer`), so the `nbytes == 0` sentinel stays correct on recycled buffers even if a caller fills lengths without calling `set_header` first. `set_header` still resets it too.

src/engines/dpdk/daqiri_dpdk_engine.cpp
cliffburdick: Fixed in fdb8500 — widened `BurstHeaderParams::nbytes` to `uint64_t` and compute the total as `static_cast<uint64_t>(ttl_len) * num_pkts`, so a large batch × frame size can't overflow the per-burst byte total. Confirmed the `custom_burst_data` static_assert still holds and pybind exposes it transparently.

Issue #175, opened 2026-06-12
Fix bare-metal link of benchmark examples to yaml-cpp


chloecrozier:
Clean CMake build on DGX Spark fails linking example binaries: `cannot find -lyaml-cpp`. Core libs build; benches do not.

Expected: `cmake --build build -j` builds all example benches with no extra flags.

Steps: `rm -rf build && cmake -S . -B build ... && cmake --build build -j` with system yaml-cpp (miniforge).

Fix: Wire yaml-cpp target in `examples/CMakeLists.txt` or document linker flags in getting-started.

Issue #174, opened 2026-06-12
RDMA client CQ errors on cross-host smoke run


chloecrozier:
Spark-to-Spark RDMA RoCE cross-host smoke (30s) logs client `CQ error` and `RETRY_EXC_ERR`. Traffic still flows (~76 Gbps); 120s soak is clean (~99 Gbps).

Expected: No CQ errors on a successful cross-host RDMA bench.

Steps: `daqiri_bench_rdma` + `daqiri_bench_rdma_tx_rx_spark_xhost.yaml`, server on RX, client on TX, `--seconds 30`. See `logs/D1-tx-v4.log`.

Fix: Check client CQ handling after connection setup on cross-host xhost.

Issue #173, opened 2026-06-12
Document routes and ARP for Spark-to-Spark RDMA


chloecrozier:
Two DGX Sparks running `daqiri_bench_rdma_tx_rx_spark_xhost.yaml` do not connect with only the bench IPs (`1.1.1.1` on TX, `2.2.2.2` on RX) on p0.

Expected: RDMA RoCE cross-host works after documented nmcli setup alone.

Steps: Run RDMA RoCE cross-host smoke per `bugs-spark-to-spark.md`; ping across bench IPs fails until `ip route` and `ip neigh` on both hosts. Soak then hits ~99 Gbps.

Fix: Add route and ARP steps to cross-host section in `docs/benchmarks/raw_benchmarking.md`.

PR #180, opened 2026-06-15
#173 #174 #175 - fix Spark cross-host RDMA bench setup and build

chloecrozier:
These bugs were found during a spark-to-spark benchmark run through.
## Issue #173 fix
- Cross-host RoCE-CM had no documented L3 step, so connect never started
- Added `scripts/setup_spark_xhost_net.sh` and a cross-host section in `docs/tutorials/system_configuration.md`
- One script run on each host installs the host route and static ARP for the direct cable

## Issue #174 fix
- xhost RDMA bench produced thousands of `RETRY_EXC`/`WR_FLUSH` CQ errors and server `Couldn't find server params` spam
- Register the listener before `rdma_bind_addr`, launch the server worker at `CONNECT_REQUEST`, demote teardown CQ statuses to DEBUG, and drain completions in `rdma_bench` before disconnect
- 30s xhost run: 0 CQ errors, matching completion counts on both sides

## Issue #175 fix
- `find_package(yaml-cpp)` half-resolved under conda and example targets failed to link
- Centralized vendored-first `yaml-cpp` resolution in `src/CMakeLists.txt` and exported `DAQIRI_YAML_TARGET` to parent scope
- Example benches link cleanly with conda active; opt out via `-DDAQIRI_PREFER_SYSTEM_YAML_CPP=ON`

## Testing
- [x] #173 ping works in both directions after running the script on each Spark
- [x] #174 30s `daqiri_bench_rdma` xhost run: 0 CQ errors, ~104 Gbps each way, matching counts
- [x] #175 `daqiri_bench_rdma`, `daqiri_bench_raw_gpudirect`, `daqiri_bench_socket` build under an activated miniforge env

## Future issues to create / bug found
- `daqiri_bench_rdma --mode client` still logs a harmless `Failed to bind for RDMA server` CRITICAL because the YAML defines both roles
- `daqiri_rdma_engine.cpp` `CONNECT_REQUEST` handler picks the last inactive slot and falls through to accept a duplicate-id client after rejecting it
- Single-host inter-port loopback (`daqiri_bench_rdma_tx_rx_spark.yaml --mode both`) fails with `UNREACHABLE` on `main`


src/engines/rdma/daqiri_rdma_engine.cpp
```diff
@@ -561,10 +574,24 @@ void RdmaEngine::rdma_thread(bool is_server, rdma_thread_params* tparams) {
...
-                           (int)wc.opcode);
+        // WR_FLUSH_ERR is the expected status for in-flight WRs when the QP
+        // moves to ERR state (peer disconnect, local shutdown). Demote to
+        // DEBUG so a clean teardown doesn't produce log spam. RETRY_EXC during
+        // shutdown is the same story: the peer's QP went away before our last
+        // outstanding WR could be acked.
+        const bool shutting_down = rdma_force_quit.load() || tparams->ready_to_exit.load();
+        const bool benign_teardown =
+            wc.status == IBV_WC_WR_FLUSH_ERR ||
+            (shutting_down && wc.status == IBV_WC_RETRY_EXC_ERR);
```
dleshchev: WR_FLUSH_ERR is demoted to DEBUG unconditionally here, while RETRY_EXC_ERR is gated on shutting_down. A flush storm following deliberate teardown is already covered because shutting_down includes the per-connection tparams->ready_to_exit flag set by the DISCONNECTED handler.

Consider gating WR_FLUSH_ERR the same way:

```cpp
const bool benign_teardown =
    shutting_down &&
    (wc.status == IBV_WC_WR_FLUSH_ERR ||
     wc.status == IBV_WC_RETRY_EXC_ERR);
```

That keeps clean teardowns quiet while still surfacing an unexpected mid-run flush, such as an abrupt peer disconnect before CM teardown is observed. If suppressing abrupt-disconnect flushes is intentional, please add a short comment explaining the asymmetry. Same issue on the TX CQ path; please mirror whichever behavior you choose.

src/engines/rdma/daqiri_rdma_engine.cpp
```diff
@@ -1376,10 +1421,11 @@ void RdmaEngine::run() {
...
             rdma_reject(params.client_id, nullptr, 0);
             continue;
-          } else {
-            DAQIRI_LOG_INFO("Server accepted connection for client ID {}",
-                              (void*)params.client_id);
           }
+
+          DAQIRI_LOG_INFO("Server accepted connection for client ID {}",
+                            (void*)params.client_id);
+          launch_server_worker(&params);
```
dleshchev, line 1438: The server worker now launches right after rdma_accept(), before RDMA_CM_EVENT_ESTABLISHED. Combined with the worker_started gate in rdma_get_server_conn_id(), the application can obtain a server conn_id and post SENDs before the connection is fully established; the handout is gated on "worker polling + rings exist," not ESTABLISHED.

For RX this is desirable because receives are primed early. For TX it opens a post-accept/pre-established window where sends can hit RNR / transport retry until the peer QP is fully ready. The benchmark masks this by priming receives first, so this is fine for #174's benchmark purpose, but it is loose for a general caller.

Consider gating the conn_id handout, or active = true, on ESTABLISHED; alternatively document that server TX must not begin until the peer is up. Fine to track as a follow-up rather than block this PR.

dleshchev (commented):
Reviewed the three fixes against #173 / #174 / #175. #173 is addressed and over-delivered with the helper script, #175's vendored-first yaml-cpp change looks sound, and #174's benchmark symptom is addressed. Docs are in sync.

I left one RDMA-engine should-fix inline around CQ error demotion, plus one follow-up discussion about server conn_id handout before ESTABLISHED.

General follow-up, pre-existing and not introduced by this PR: in the CONNECT_REQUEST handler, the slot-selection loop has no break after queue_idx = i, so it selects the last inactive slot rather than the first; and the duplicate-client-id branch rdma_reject()s then breaks the loop but falls through to rdma_accept() the rejected client anyway. You already called both out under "Future issues" in the PR description; please file a tracking issue so they are not lost.

dleshchev:
One remaining should-fix from the update: `src/engines/rdma/daqiri_rdma_engine.cpp:1391` still selects a free server slot using `!active`, but `active` is only set later when `rdma_get_server_conn_id()` hands the conn_id to the app. A slot with `client_id != nullptr && active == false` is already occupied by an accepted-but-not-yet-claimed connection; a second CONNECT_REQUEST before the app polls can reuse that same slot, overwrite `params.client_id`, and leave the first worker running against stale/mismatched params/rings.

The duplicate-client reject and first-slot break fixes are good, but this free-slot test should key off occupancy instead, e.g. choose only `client_id == nullptr` slots, or add a separate occupied/assigned state distinct from app-visible `active`.

Issue #167, opened 2026-06-12
[BUG] Dead drop jump field and no flow smoke test


chloecrozier:
Found while working on PR 161 (#128).

**Issue:** DropTrafficConfig::jump is unused after ensure_eth_jump_rule(). There is no documented hardware check for NIC flow programming.

**Expected:** No dead fields in the drop-traffic path. A short manual test confirms good configs init and bad queue IDs fail.

**Fix:** Remove DropTrafficConfig::jump and related null assignments. Add manual smoke steps to the benchmarks or contributing docs.

Issue #166, opened 2026-06-12
[BUG] Flex flows use global handle and allow mixed configs


chloecrozier:
Found while working on PR 161 (#128).

**Issue:** create_flex_flow_rule() uses one static handle (first port wins). validate_config allows standard and flex-item RX flows on the same interface.

**Expected:** Each flex_item_id gets its own handle. Mixed flow types on one interface are rejected at config time.

**Fix:** Replace the static singleton with a per flex_item_id map. Add a validate_config check that errors when an interface has both standard and flex-item flows.

PR #181, opened 2026-06-15
#165 #166 #167 - Complete DPDK flow rule lifecycle

### LINKED ISSUE #165: [BUG] DPDK RX and TX flows not destroyed on shutdown
Found while working on PR 161 (#128).

**Issue:** RX steering and tx_eth_src NIC rules are created at init but never destroyed on exit. Only ETH jump rules are torn down.

**Expected:** All flow rules created by DAQIRI are destroyed in shutdown and failed-init cleanup.

**Fix:** Track rte_flow* handles from add_flow(), add_flex_item_flow(), and add_modify_flow_set(), then call rte_flow_destroy() in shutdown() and the init-failure path.

chloecrozier:
Closes #165, #166, #167

Issue #165 fix
- RX steering, send-to-kernel fallback, and tx_eth_src NIC rules were created at init but never destroyed on exit.
- All rte_flow handles are now tracked and destroyed in shutdown, init-failure cleanup, and the destructor.
- Flex-item handles are released via rte_flow_flex_item_release before ETH jump rules are torn down.

Issue #166 fix
- create_flex_flow_rule() used one static handle so only the first flex_item_id worked.
- Flex items are now created per (port, flex_item_id) with lookup by FlexItemConfig.id_ on the correct interface.
- Mixed standard and flex-item flows on one interface were already rejected in validate_config (#129); no change needed.

Issue #167 fix
- DropTrafficConfig::jump was dead after ensure_eth_jump_rule() took ownership.
- Removed the unused jump field and its null assignments.
- Added manual flow programming smoke steps to raw_benchmarking.md.

testing:
- [x] #165 Container build of daqiri_dpdk succeeds
- [x] #165 Flow destroy wired into shutdown, EalCleanupGuard, and destructor
- [x] #166 Invalid action.id fails at validate_config before NIC init
- [x] #166 Invalid flex_item_id fails at validate_config before NIC init
- [x] #167 Dead jump field removed; smoke doc steps verified
- [x] Build smoke: daqiri_bench_raw_sw_loopback.yaml
- [x] NIC smoke: cross-host raw tx/rx on p0 (169.254.95.47 → 169.254.100.253), RX flow + tx_eth_src programmed, 24M pkts matched


dleshchev (commented):
Reviewed #165 / #166 / #167. The flow lifecycle fix is substantially in place: RX, fallback, TX offload, and flex-item resources are tracked and destroyed with the intended ordering, flex handles are now scoped per (port, flex_item_id), and the dead drop jump field is removed. Docs are in sync.

One should-fix validation gap inline below; otherwise this looks good.

src/engines/dpdk/daqiri_dpdk_engine.cpp
```diff
@@ -3455,6 +3525,20 @@ bool DpdkEngine::validate_config() const {
       flow_to_queue.emplace(flow.id_, flow.action_.id_);
       if (flow.match_.type_ == FlowMatchType::FLEX_ITEM) {
         has_flex_item_flows = true;
+        const uint16_t flex_item_id = flow.match_.flex_item_match_.flex_item_id_;
```
dleshchev, line 3618: validate_config() now checks that a flex flow references an existing flex_item_id, which is good, and it already rejects duplicate flow IDs. However, it does not reject duplicate id_ values within intf.rx_.flex_items_ itself.

Both this check and find_flex_item_config() return the first matching ID, so two flex_items sharing an ID with different parser settings would silently bind to the first parser. Since flex_item_id is now the config contract for lookup, please add a per-interface uniqueness check, for example:

```cpp
std::unordered_set<uint16_t> flex_ids;
for (const auto& item : intf.rx_.flex_items_) {
  if (!flex_ids.insert(item.id_).second) {
    DAQIRI_LOG_ERROR("Duplicate flex item id {} on interface '{}'",
                     item.id_,
                     intf.name_);
    return false;
  }
}
```

Low severity because no shipped YAML uses flex_items, but it is the same kind of malformed config the rest of this function now guards against.

PR #202, opened 2026-06-22
Validate packet burst bounds before writes

fallintoplace:
## Summary
- add a shared burst validation helper for packet counts, segment indexes, packet indexes, payload lengths, packet lengths, and backing capacities
- guard public burst accessors, payload setters, packet-length setters, TX burst acquisition, send, and release paths across DPDK, socket, RDMA, and ibverbs engines
- reject oversized or malformed packet metadata before memcpy, length assignment, or segment access

## Testing
- `git diff --check`
- `cmake -S . -B build-verify -G Ninja -DDAQIRI_ENGINE="" -DDAQIRI_BUILD_EXAMPLES=OFF -DDAQIRI_BUILD_PYTHON=OFF` (fails locally before targets: CUDA toolkit/nvcc is not installed)


cliffburdick:
This PR appears to add a significant amount of checks that are per-packet or per-segment, which quickly has a large overhead for small packet sizes. Can you characterize the performance loss from having this, and if it's noticeable, put it behind an compilation flag?

PR #208, opened 2026-06-30
#207: Fix TX mempool exhaustion deadlock in tx_core_worker

### LINKED ISSUE #207: [BUG] TX mempool exhaustion stall in DPDK tx_core_worker
## Describe the bug

When the TX ring drains, `tx_core_worker` idles without polling NIC completions. Completed mbufs are never reclaimed, the mempool free count stays below the `2 × batch_size` gate in `is_tx_burst_available()`, and `send_tx_burst` is permanently blocked.

### Root Cause

`send_tx_burst` (app thread) gates new submissions on mempool availability:

```cpp
// is_tx_burst_available()
if (rte_mempool_avail_count(q->pools[seg]) < burst->hdr.hdr.num_pkts * 2) {
  return false;  // → caller gets NO_FREE_PACKET_BUFFERS
}
```

`tx_core_worker` (TX lcore) only reclaims mbufs via `rte_eth_tx_burst()`, which is only called when the ring has work. When the ring is empty, the idle path does nothing:

```cpp
// tx_core_worker main loop
while (!force_quit.load()) {
  if (rte_ring_dequeue(tparams->ring, reinterpret_cast<void**>(&msg)) != 0) {
    continue;  // ← no completion polling, mbufs never reclaimed
  }
  // ... rte_eth_tx_burst() only reached below ...
}
```

Once the pool is exhausted, neither side can make progress.

### Example (num_bufs=12288, batch_size=4096, num_tx_desc=8192)

Gate threshold = `batch_size × 2` = 8192. Both bursts fit in the NIC descriptor ring (8192 descriptors), so `tx_core_worker` finishes and enters idle before the NIC completes any transmission.

| Step | Action | Pool free | Ring | NIC in-flight |
|------|--------|-----------|------|---------------|
| 1 | initial state | 12288 | 0 | 0 |
| 2 | `send_tx_burst`: burst 1 (4096) | 8192 | 1 | 0 |
| 3 | `send_tx_burst`: burst 2 (4096) | 4096 | 2 | 0 |
| 4 | `send_tx_burst`: 4096 < 8192 gate | **BLOCKED** | 2 | 0 |
| 5 | `tx_core_worker`: tx_burst(4096) | 4096 | 1 | 4096 |
| 6 | `tx_core_worker`: tx_burst(4096) | 4096 | 0 | 8192 |
| 7 | `tx_core_worker`: ring empty → `continue` | 4096 | 0 | 8192 |
| 8 | NIC completes all 8192 packets | 4096 | 0 | 8192 (**never reclaimed**) |

After step 8 the pool is stuck at 4096 free < 8192 gate. `send_tx_burst` stays blocked, `tx_core_worker` spins on empty ring, and the 8192 completed mbufs are never returned to the pool.

## Steps to reproduce

Save the following script as `repro.sh` and run it inside the daqiri container (set `NIC_ADDR`, `IP_SRC`, `IP_DST` as needed):

```bash
docker run --rm --privileged --net=host \
  --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all \
  -v /dev/hugepages:/dev/hugepages \
  -v "$(pwd)/repro.sh:/workspace/repro.sh" \
  daqiri:local /workspace/repro.sh
```

`repro.sh`:

```bash
#!/usr/bin/env bash
set -euo pipefail

NIC_ADDR="${NIC_ADDR:-0005:03:00.0}"
IP_SRC="${IP_SRC:-10.0.0.1}"
IP_DST="${IP_DST:-10.0.0.2}"
MIN_PACKETS="${MIN_PACKETS:-1000000}"

BENCH=""
for d in ./build/examples /opt/daqiri/bin; do
  [[ -x "$d/daqiri_bench_raw_gpudirect" ]] && BENCH="$d/daqiri_bench_raw_gpudirect" && break
done
[[ -z "$BENCH" ]] && { echo "daqiri_bench_raw_gpudirect not found"; exit 1; }

CONFIG="/tmp/repro_tx_stall_$$.yaml"
trap 'rm -f "$CONFIG"' EXIT

cat > "$CONFIG" <<EOF
%YAML 1.2
---
daqiri:
  cfg:
    version: 1
    stream_type: "raw"
    master_core: 0
    debug: false
    log_level: "warn"
    loopback: ""
    memory_regions:
    - name: "Data_TX"
      kind: "host_pinned"
      affinity: 0
      num_bufs: 12288
      buf_size: 1064
    interfaces:
    - name: "tx_port"
      address: "${NIC_ADDR}"
      tx:
        queues:
        - name: "tx_q_0"
          id: 0
          batch_size: 4096
          cpu_core: 3
          memory_regions:
            - "Data_TX"
          offloads:
            - "tx_eth_src"
bench_tx:
  interface_name: "tx_port"
  cpu_core: 4
  batch_size: 4096
  payload_size: 1000
  header_size: 64
  eth_dst_addr: "ff:ff:ff:ff:ff:ff"
  ip_src_addr: "${IP_SRC}"
  ip_dst_addr: "${IP_DST}"
  udp_src_port: 4096
  udp_dst_port: 4096
EOF

OUTPUT=$(timeout --signal=KILL 30 "$BENCH" "$CONFIG" --seconds 10 2>&1) && rc=0 || rc=$?
echo "$OUTPUT"

if [[ $rc -eq 137 ]]; then
  echo -e "\nFAIL: killed after 30s — TX deadlock."
  exit 1
fi

TX_PACKETS=$(grep -oP 'packets=\K[0-9]+' <<< "$OUTPUT" | head -1)
: "${TX_PACKETS:=0}"

echo ""
echo "Transmitted: $TX_PACKETS packets (minimum: $MIN_PACKETS)"

if (( TX_PACKETS < MIN_PACKETS )); then
  echo -e "\nFAIL: throughput collapse — only $TX_PACKETS packets in 10s (need $MIN_PACKETS)."
  exit 1
fi

echo -e "\nPASS"
```

Output (look for the `TX complete` line):

```
TX complete: interface=tx_port queue=0 packets=~20480 bytes=21790720 bursts=5 seconds=10.0065
```

## Expected behavior

The benchmark should sustain >10 Mpps for the full 10-second run (>100M packets total).

## Actual behavior

Throughput collapses after 2 bursts — only ~20K packets sent in 10 seconds (~2 Kpps).

## Environment overview

- Environment location: Docker
- Method of DAQIRI install: source (container build via `./scripts/build-container.sh`)

## Environment details

- OS: Ubuntu 24.04 (container base: `nvcr.io/nvidia/cuda:13.1.0-devel-ubuntu24.04`)
- DAQIRI version: main branch
- DPDK version: 25.11
- Hardware: IGX Dev Kit, ConnectX-7 (mlx5 PMD)

lobolanja:
Fixes #207

## Problem

`tx_core_worker` never polls NIC TX completions while idling (empty ring), so mbufs are never reclaimed and `is_tx_burst_available()` blocks the producer permanently. Throughput collapses from >10 Mpps to ~2 Kpps.

## Fix

Poll completions in the `tx_core_worker` idle path so mbufs are returned to the pool even when no new work is queued.

### Why the DPDK patch is needed

DPDK provides `rte_eth_tx_done_cleanup()` as the standard API for explicitly reclaiming completed TX mbufs without sending new packets. However, the mlx5 PMD in DPDK 25.11 does **not** implement the `tx_done_cleanup` callback — calling it returns `-ENOTSUP`.

The only ways to trigger mlx5 completion processing (`mlx5_tx_handle_completion`) are:

| Method | API | Problem |
|--------|-----|---------|
| Send packets | `rte_eth_tx_burst()` | Requires work in the ring — useless when idle |
| Check descriptor | `rte_eth_tx_descriptor_status()` | Calls `mlx5_tx_handle_completion()` as a **side effect**, but that's not the function's contract — it is meant to report descriptor state, not reclaim mbufs. Future DPDK versions could change this behavior. |
| Done cleanup | `rte_eth_tx_done_cleanup()` | Correct API, but returns `-ENOTSUP` on mlx5 |

The patch adds a trivial `mlx5_tx_done_cleanup()` (~20 lines) that wraps the **existing** `mlx5_tx_handle_completion()` — no new completion logic is introduced. It simply exposes the internal function through the official DPDK callback so `rte_eth_tx_done_cleanup()` works as intended.

Without this patch, the only alternative is the `rte_eth_tx_descriptor_status` side-effect workaround (commit 1), which is fragile and semantically incorrect.

## Commits

| Commit | Description |
|--------|-------------|
| `9f62fe2` | **Workaround** — call `rte_eth_tx_descriptor_status(port, queue, 0)` in idle path. On mlx5 this triggers `mlx5_tx_handle_completion()` as a side effect. |
| `19f5329` | **DPDK patch** — add `mlx5_tx_done_cleanup()` to mlx5 PMD as a wrapper around `mlx5_tx_handle_completion`. Registered in `mlx5_dev_ops` and `mlx5_dev_ops_isolate`. |
| `a7354dc` | **Use proper API** — replace descriptor_status workaround with `rte_eth_tx_done_cleanup(port, queue, 0)`. |

## Reproduction

See the self-contained test script in issue #207.

```bash
# Without fix (demonstrates the bug):
# → ~20K packets in 10s → FAIL

# With fix:
# → >10M packets in 10s → PASS
```

## Testing

- Container build: clean, no warnings
- Throughput test PASS after commit 1 (workaround) and after commit 3 (final fix)

### DIFF
--- a/dpdk_patches/mlx5_tx_done_cleanup.patch
+++ b/dpdk_patches/mlx5_tx_done_cleanup.patch
@@ -1,0 +1,59 @@
+diff --git a/drivers/net/mlx5/mlx5_tx.h b/drivers/net/mlx5/mlx5_tx.h
+--- a/drivers/net/mlx5/mlx5_tx.h
++++ b/drivers/net/mlx5/mlx5_tx.h
+@@ -237,6 +237,7 @@
+ void mlx5_tx_handle_completion(struct mlx5_txq_data *__rte_restrict txq,
+ 			       unsigned int olx __rte_unused);
+ int mlx5_tx_descriptor_status(void *tx_queue, uint16_t offset);
++int mlx5_tx_done_cleanup(void *tx_queue, uint32_t free_cnt);
+ void mlx5_txq_info_get(struct rte_eth_dev *dev, uint16_t queue_id,
+ 		       struct rte_eth_txq_info *qinfo);
+ int mlx5_tx_burst_mode_get(struct rte_eth_dev *dev, uint16_t tx_queue_id,
+diff --git a/drivers/net/mlx5/mlx5_tx.c b/drivers/net/mlx5/mlx5_tx.c
+--- a/drivers/net/mlx5/mlx5_tx.c
++++ b/drivers/net/mlx5/mlx5_tx.c
+@@ -288,5 +288,27 @@ mlx5_tx_descriptor_status(void *tx_queue, uint16_t offset)
+ }
+ 
++/**
++ * DPDK callback to free consumed TX mbufs.
++ *
++ * @param tx_queue
++ *   The TX queue.
++ * @param[in] free_cnt
++ *   Maximum number of packets to free. Use 0 to indicate all possible packets
++ *   should be freed. Note that a packet may be using multiple mbufs.
++ *
++ * @return
++ *   Number of packets freed, or negative errno on error.
++ */
++int
++mlx5_tx_done_cleanup(void *tx_queue, uint32_t free_cnt __rte_unused)
++{
++	struct mlx5_txq_data *__rte_restrict txq = tx_queue;
++	uint16_t tail_before = txq->elts_tail;
++
++	mlx5_tx_handle_completion(txq, 0);
++	return (uint16_t)(txq->elts_tail - tail_before);
++}
++
+ /*
+  * Array of declared and compiled Tx burst function and corresponding
+  * supported offloads set. The array is used to select the Tx burst
+diff --git a/drivers/net/mlx5/mlx5.c b/drivers/net/mlx5/mlx5.c
+--- a/drivers/net/mlx5/mlx5.c
++++ b/drivers/net/mlx5/mlx5.c
+@@ -2585,5 +2585,6 @@ const struct eth_dev_ops mlx5_dev_ops = {
+ 	.rxq_info_get = mlx5_rxq_info_get,
+ 	.txq_info_get = mlx5_txq_info_get,
++	.tx_done_cleanup = mlx5_tx_done_cleanup,
+ 	.rx_burst_mode_get = mlx5_rx_burst_mode_get,
+ 	.tx_burst_mode_get = mlx5_tx_burst_mode_get,
+ 	.rx_queue_intr_enable = mlx5_rx_intr_enable,
+@@ -2680,5 +2681,6 @@ const struct eth_dev_ops mlx5_dev_ops_isolate = {
+ 	.rxq_info_get = mlx5_rxq_info_get,
+ 	.txq_info_get = mlx5_txq_info_get,
++	.tx_done_cleanup = mlx5_tx_done_cleanup,
+ 	.rx_burst_mode_get = mlx5_rx_burst_mode_get,
+ 	.tx_burst_mode_get = mlx5_tx_burst_mode_get,
+ 	.rx_queue_intr_enable = mlx5_rx_intr_enable,
--- a/src/engines/dpdk/daqiri_dpdk_engine.cpp
+++ b/src/engines/dpdk/daqiri_dpdk_engine.cpp
@@ -6110,7 +6110,16 @@ int DpdkEngine::tx_core_worker(void* arg) {
                     (void*)tparams->ring);
 
   while (!force_quit.load()) {
-    if (rte_ring_dequeue(tparams->ring, reinterpret_cast<void**>(&msg)) != 0) { continue; }
+    if (rte_ring_dequeue(tparams->ring, reinterpret_cast<void**>(&msg)) != 0) {
+      // Poll TX completions while idle so the PMD reclaims mbufs back to the
+      // mempool.  Without this, is_tx_burst_available() can permanently gate
+      // new submissions once the pool drops below the 2×batch threshold,
+      // because rte_eth_tx_burst() — the only other reclaim path — is never
+      // called when the ring is empty.  On mlx5 this dispatches to
+      // mlx5_tx_descriptor_status() → mlx5_tx_handle_completion().
+      (void)rte_eth_tx_descriptor_status(tparams->port, tparams->queue, 0);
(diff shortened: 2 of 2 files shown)


cliffburdick:
Hi @lobolanja, this appears to be a patch for DPDK. Should this be submitted to that project?

lobolanja:
Hi @cliffburdick, Yes, the DPDK patch (mlx5_tx_done_cleanup) is intended for upstream submission as well. I started here because I wanted to validate the approach with the DAQIRI team first: I wasn't fully confident this is a genuine gap rather than a misuse on mi side (e.g., maybe there's a different intended pattern for idle-path mbuf reclaim that I'm missing).

I'm also happy to submit the mlx5 patch upstream to DPDK via their mailing list process, though that's a separate and longer cycle. In the meantime, carrying the patch in daqiri's dpdk_patches/ keeps things working without waiting on upstream.

If the fix looks correct to you, I'll go ahead and open a patch on dpdk.org in parallel.

cliffburdick:
Hi @lobolanja, we're trying to migrate most users to our ibverbs backend since it performs better than the DPDK backend. This means we're also trying to deprecate the patches. I don't know enough about the mlx5 tx freeing to know whether this is needed, so perhaps you could submit it there and see if it gets accepted. In my experience things get reviewed within a few days of you submitting the patches, and it would at least be on the dev branch of DPDK.

lobolanja:
@cliffburdick 

That makes sense, thanks for the context on the ibverbs migration.

In that case, would you be okay with keeping just commit 9f62fe2 (the rte_eth_tx_descriptor_status workaround)? It fixes the deadlock without adding any DPDK patch to maintain. I can rebase the PR to drop commits 2 and 3.

In parallel, I'll submit the tx_done_cleanup patch upstream to DPDK. Once it lands there, the workaround here can be swapped for the proper API in a follow-up, or it may become a non-issue if the engine moves to ibverbs by then.

cliffburdick:
Yes, I think that's reasonable

PR #223, opened 2026-07-10
#15 - Clean up of RoCE bench issues

RamyaGuru:
Clean up of various RoCE bench issues on DGX Spark:

- **Unidirectional standardization** — RoCE was the only transport running
  bidirectionally; make it one-way by default (matching DPDK and sockets) so the
  numbers are apples-to-apples, and re-measure the perf report.
- **Idling in the workload bench** — the RoCE workload receive path blocked on the
  GPU every message; recycle recv buffers with CUDA events so the receiver keeps
  reaping completions instead.
- **Fixed GEMM size for the workload bench** — pin the GEMM dimension (and add a
  matching FFT-length knob) so the FLOP count per call is constant across the
  message-size sweep.
- **Spark netns port auto-detect** — harden the wire-loopback setup to group ports
  by `phys_port_name`, robust to PCIe re-enumeration.


docs/benchmarks/performance-dgx-spark.md
```diff
@@ -27,27 +31,27 @@ loopback). The exact commands are collected under [Reproduce](#reproduce) below.
...
+Each transport at its best-case **operation size**. Raw/RoCE are
 single-stream; socket TCP/UDP scale with the number of client/server pairs, so the
 four-pair aggregate is shown.
 
 | Stream / Protocol | Best case | Throughput | Drops | Notes |
 | ----------------- | --------- | ---------: | ----- | ----- |
-| Raw Ethernet / GPUDirect | 4 KB packet | **105.5 ±0.9 Gb/s** | 0 | 98.5 Gb/s single-queue at the 8 KB native shape |
-| Socket / RoCE (SEND) | 8 MB message | **102.2 ±0.3 Gb/s** | 0 | Single QP, batch 1 |
+| Raw Ethernet / GPUDirect | 4 KB packet | **105.5 ±0.9 Gb/s** | 0 | 98.5 Gb/s single-queue at 8 KB |
+| Socket / RoCE (SEND) | 64 KB message | **106.0 ±0.1 Gb/s** | 0 | Single QP, batch 1 |
```
cliffburdick: I think we should just artificially say this is 100Gbps since it could be confusing to people without a footnote. The Spark networking is very strange so 106Gbps could look *bad* in some cases if people think it's a 200G link.

docs/benchmarks/performance-dgx-spark.md
```diff
@@ -143,36 +147,37 @@ saturate the wire; the smallest messages are bound by per-operation software
...
 
-| Core                | Busy% | Note                                            |
-| ------------------- | ----: | ----------------------------------------------- |
-| Master (CPU 8)      |  5.9% | Orchestration only                              |
-| Client TX (CPU 17)  | 74.9% | Post-and-poll spin; rate-independent            |
-| Server RX (CPU 18)  |  0.0% | HCA writes straight to memory; CPU uninvolved   |
+| Core                 | Busy% | Note                                            |
+| -------------------- | ----: | ----------------------------------------------- |
+| Master (CPU 8)       |  0.7% | Orchestration only                              |
+| Client TX (CPU 17)   | 74.8% | Post-and-poll spin; rate-independent            |
```
cliffburdick: What does rate-independent mean here?

examples/bench_workload.h
```diff
@@ -32,11 +33,42 @@ enum class BenchWorkload { None, Fft, Gemm, GemmFp16 };
...
+// FIXED so the FLOP count per call (2*n^3) is constant as the I/O unit (message /
+// burst window) is swept -- this isolates pipelining depth from problem size in
+// the RoCE-vs-raw comparison. The compute working set is n*n*elem_size, derived
+// entirely from n; the assembled input buffer must be at least that large.
+// Returns 1024 (the default) if unset. Mirrors parse_workload's stride.
+int parse_workload_gemm_dim(int argc, char** argv);
+
+// Parse "--workload-fft-len N" from argv: the 1-D C2C transform length. The burst
+// working set is fanned out across as many batched length-N transforms as fit, so
+// N sets the per-transform cost (~5*N*log2(N) flops) while the batch count tracks
```
cliffburdick: That equation only works for powers of two, so we may want to restrict it or remove that comment

docs/benchmarks/performance-dgx-spark.md
```diff
@@ -292,74 +294,24 @@ across the run. Throughput is mean ± std over 3 reps, 30 s each.
...
-(`gemm_fp16`, single rep per point. Raw cells use the nearest whole-packet window to the
-batch size; RoCE caps the batch at the 8 MB message.)
+Fixed **n=1024**, one GEMM (or a length-1024 batched FFT) per received unit. DPDK runs
+at an **8 KB payload** (~8 MB reorder window, 1024 packets × 8000 B), matched to RoCE's
+**8 MB message** so the GPU working set and per-unit compute are the same on both.
+3 reps, 30 s each, GPU SM% from `nvidia-smi dmon`; 0 drops on every cell.
+
+| Workload | DPDK (Raw / GPUDirect) | RoCE (RC) |
+| -------- | ---------------------: | --------: |
+| none (baseline) | 98.4 ±0.2  | 108.0 ±0.1 |
```
cliffburdick: This one is surprising to me. Do we know why the packet reception went down by 5% when a workload was enabled? I would think given the memory bandwidth it would be unaffected.

cliffburdick (approved):

docs/benchmarks/performance-dgx-spark.md
RamyaGuru: No, this was it :/ I had grabbed a second QSFP from the office that turned out to be 200 instead of 100 Gps, and think I must have been running with that cable for some time. I'm so sorry for the confusion! I refreshed all numbers in this doc with a vetted 100G link. It reproduced most of the old numbers, but this time saturated at wire rate properly instead of giving > 100 Gbps throughput values.
I think the old report was capping ~108 Gbps on the 200G link because I was measuring unidirectional traffic only p0 --> p1?

docs/benchmarks/performance-dgx-spark.md
RamyaGuru: I think meaning that the CPU % for polling CPUs is not dependent on the data rate on the wire? But I **removed that phrase** because I don't believe it was something I actually tested for.

examples/bench_workload.h
RamyaGuru: Thanks. Removed comment for now.

docs/benchmarks/performance-dgx-spark.md
RamyaGuru: I don't have a clear explanation yet. I removed speculative explainer text in the report. I think another data point on IGX or RTX Pro server would help me see if it has something to do with unified memory on the Spark that I'm not handling right?

docs/benchmarks/performance-dgx-spark.md
```diff
@@ -44,8 +44,8 @@ just under that ceiling.
 | ----------------- | --------- | ---------: | ----- | ----- |
 | Raw Ethernet / GPUDirect | 8 KB packet | **98.8 ±0.1 Gb/s** | 0 | Flat ~98.7 across 4–8 KB, all batch sizes |
 | Socket / RoCE (SEND) | 64 KB message | **97.6 ±0.1 Gb/s** | 0 | Single QP, batch 1 |
-| Socket / TCP | 1 MiB × 4 pairs | **90.6 ±1.5 Gb/s** | ~0 | Flow-controlled (App TX = App RX) |
-| Socket / UDP | 8 KB × 4 pairs | **28.5 ±0.6 Gb/s** | ~54% loss | Receiver goodput; unpaced sender |
+| Socket / TCP | 8 KB × 4 pairs | **87.3 ±2.2 Gb/s** | ~0 | Flow-controlled (App TX = App RX) |
```
cliffburdick, line 47: These numbers are still backwards to me. We probably should investigate at some point

Issue #220, opened 2026-07-09
[FEA] Add eMPW to Tx


cliffburdick:
Add eMPW for more efficiency on TX

cliffburdick:
Closed by #221

PR #224, opened 2026-07-14
#220 - Size ibverbs TX QPs for actual segments

cliffburdick:


RamyaGuru:
Validated on IGX Thor DevKit. ibverbs engine with num_bufs = 12288 now works

RamyaGuru (approved):

Issue #206, opened 2026-06-26
[DOC] Finish README and website positioning polish


dleshchev:
## Report incorrect documentation

**Location of incorrect documentation**

- `README.md`
- `docs/images/logo.svg`
- `mkdocs.yml`
- `docs/benchmarks/benchmarks.md`
- Landing page snippets under `docs/landing/`

**Describe the problems or issues found in the documentation**

A few README and website positioning items need cleanup so the first-time reader gets a clearer picture of DAQIRI's hardware requirements, engine choices, peak-performance path, and benchmark results.

Current issues:
- The README logo artwork has the daiquiri glass and compute/server element on different visual baselines.
- README does not have a concise engine summary that says:
  - DPDK is the default raw Ethernet engine.
  - raw `ibverbs` is an opt-in engine for supported Mellanox/mlx5 NICs.
  - Linux TCP/UDP sockets are always available.
- The README and website should distinguish Linux TCP/UDP sockets from RoCE more clearly. TCP/UDP sockets do not require an NVIDIA NIC; RoCE is configured through the socket stream model but still requires suitable RDMA / ConnectX hardware.
- The README and website should state that peak-performance benchmarking and production datapaths should use the C++ API.
- Benchmark numbers are not visible enough in the website landing flow.
- The docs page is still labeled "Concepts" in navigation even though the page describes itself as the DAQIRI glossary.
- The Benchmarking page decision-tree image is too large on first load.

**Steps taken to verify documentation is incorrect**

- Checked `origin/main` README and docs navigation.
- Checked current landing-page includes in `docs/index.md`.
- Checked recent PR history for website/news handling.
- Checked existing benchmark docs and open benchmark platform issues.

**Suggested fix for documentation**

- Adjust `docs/images/logo.svg` so the daiquiri glass and compute/server element sit on the same horizontal plane.
- Add a short README "Engines" section using conservative wording:
  - raw Ethernet defaults to `dpdk`;
  - raw `ibverbs` is opt-in for supported mlx5 hardware;
  - Linux TCP/UDP sockets are always built in;
  - RoCE uses the socket stream configuration model but requires RDMA-capable hardware.
- Add a short C++ peak-performance note to README and the website/API entry points.
- Add a compact benchmark teaser to the landing page with available measured numbers and links to detailed benchmark pages.
- Rename the user-facing docs nav label from "Concepts" to "Glossary", if that is the intended terminology.
[...]

PR #227, opened 2026-07-15
#206 - Docs website cleanup: nav, mobile toggle, header, and prose

chloecrozier:
## Summary

Docs cleanup for the website (issue #206). Fixes two visible UI bugs, removes and obvious ai-speak, and keeps the wordage simple and clear
## Changes

- **Fix mobile light/dark toggle.** The palette had two entries both bound to `prefers-color-scheme`, so tapping the button only mirrored the OS setting and appeared to do nothing. Replaced it with a simple two-state toggle that defaults to dark and switches dark, light, dark
- **Fix landing header title on scroll.** The site title and page title were identical, so Material slid the same text sideways when scrolling from the top. The landing header now shows one static title
- **Remove duplicate Previous/Next nav.** Pages had hand-written Previous/Next links that duplicated (and disagreed with) the theme's built-in footer nav. Kept the theme footer as the single source
- **Shorten the site title.** Uses a colon ("DAQIRI:") instead of a dash. It is shorter and reads more cleanly, including for screen readers
- **Drop em dashes and prose semicolons across the docs.** Rewrote them as commas, periods, or parentheses
- Shorter clauses are easier to read and are announced more predictably by screen readers. Code blocks are untouched
- **Plainer wording.** Toned down a few marketing-style sentences on the landing page.
- **Small grammar fix.** "a NVIDIA" to "an NVIDIA" for consistency

## Test plan

- [x] `mkdocs build --strict` passes (only the pre-existing packet-diagram image warnings, which CI generates)
- [x] `scripts/check_html_links.py site` passes
- [x] `scripts/check_doc_refs.py` passes
- [ ] Reviewer: UI looks good locally

### DIFF
--- a/docs/api-reference/configuration.md
+++ b/docs/api-reference/configuration.md
@@ -54,7 +54,7 @@ These settings apply globally to both TX and RX:
 
 ## Memory Regions
 
-`memory_regions:` — List of regions where packet buffers are stored. The number of regions
+`memory_regions:` List of regions where packet buffers are stored. The number of regions
 and their `kind` determines the receive mode (CPU-only, header-data split, or batched GPU).
 
 - **`name`**: Memory region name. Referenced by queue configurations.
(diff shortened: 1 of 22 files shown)


dleshchev:
The landing header override hides the second `.md-header__topic`, but Material still toggles `.md-header__title--active` after scrolling past the H1. That active rule also sets the first topic to `pointer-events: none` and `z-index: -1`, while this override only restores `opacity`/`transform`. Could you also reset `pointer-events` and `z-index` here, or otherwise disable the active title state on the landing page?

chloecrozier:
@dleshchev  I addressed your comment!

PR #243, opened 2026-08-05
#73 - add ResNet-50 inference example with cross-host ingest benchmarks

### LINKED ISSUE #73: [FEA] Add TensorRT inference example: packet reorder → ResNet-50 feature extraction → PCA visualization
## Summary

Add a new example application that demonstrates an end-to-end DAQIRI + TensorRT inference pipeline. The example receives packets via GPUDirect, reassembles them into input tensors (images or time-series signals), runs ResNet-50 feature extraction through a TensorRT engine, and produces a live PCA plot of the accumulated latent-space vectors.

This fills a gap in the current examples: all existing benchmarks (`raw_gpudirect`, `raw_hds`, `raw_reorder_*`, `rdma`, `socket`) focus on network I/O throughput. There is no example showing how received data flows into a GPU inference workload — the use case that the `BASE_IMAGE=torch` container build path was designed to support.

## Motivation

- Users frequently ask how to connect DAQIRI packet ingestion to a downstream AI/ML pipeline.
- The README already mentions TensorRT inference workflows but provides no working example.
- A concrete, runnable example lowers the barrier for data-acquisition + inference adoption on DGX Spark, IGX Thor, and RTX Pro Server hardware.

### Data flow

```
NIC → DAQIRI RX (GPUDirect, HDS or batched-GPU) → packet reorder (sequence number)
    → reassemble into 224×224×3 image tensor
    → TensorRT ResNet-50 engine (feature extraction, output = 2048-d vector)
    → accumulate latent vectors on host
    → incremental PCA → print PC1, PC2 to stdout
```

### Input mode

Reorder incoming packets by sequence number and reconstruct a 224×224×3 image (or batch of images) suitable for ResNet-50 input. This models a camera or imaging-sensor ingest workflow where raw pixel data arrives as UDP packets and must be reassembled in order before inference.

### TensorRT integration

- Build a ResNet-50 TensorRT engine from an ONNX model at first run (cache the serialized engine for subsequent runs).
- Use the engine for feature extraction only — remove the final classification head and output the 2048-dimensional average-pool vector.
- Batch size should be configurable in the YAML (default 8).

### PCA output

- Accumulate feature vectors in a host-side buffer.
- After every N batches (configurable), run incremental PCA (2 components) and print the PC1 and PC2 values to stdout.
- Fully headless — no graphics dependencies, no plot files. Users can pipe the output to a CSV or their own visualization tool if desired.

### New files

| File | Description |
|------|-------------|
| `examples/tensorrt_inference_example.cpp` | Main application source |
| `examples/daqiri_example_tensorrt_inference.yaml` | YAML config (loopback-friendly defaults) |
| `examples/tensorrt_inference_utils.h` | TensorRT engine build/load helpers |
| `examples/tensorrt_inference/README.md` | Step-by-step build and run instructions |

### README and build instructions

- Add `examples/tensorrt_inference/README.md` with step-by-step instructions covering:
  - How to build the `torch` container (`BASE_IMAGE=torch scripts/build-container.sh`)
  - How to download or export the ResNet-50 ONNX model
  - Manual compile commands (`g++` / `nvcc` with the correct TensorRT and DAQIRI include/link flags)
  - How to run the example with the provided YAML config
  - Expected terminal output and how to interpret the PC1/PC2 values
- The example is **not** wired into the main `examples/CMakeLists.txt` — it lives as a standalone recipe that users compile manually inside the `torch` container.

### Documentation updates (per docs-sync rules)

- `docs/tutorials/benchmarking_examples.md` — add a section for the TensorRT inference example.
- `docs/tutorials/configuration-walkthrough.md` — add a leaf in the "Choosing an example config" decision tree for the new YAML configs.
- `CLAUDE.md` — add the new executable and configs to the Benchmarks table.
- `README.md` — mention the inference example in the Features list or a new "AI/ML Integration" bullet.

## Acceptance criteria

- [ ] `tensorrt_inference_example` compiles successfully following the README instructions inside the `torch` container.
- [ ] The example runs end-to-end in software loopback mode (`daqiri_bench_raw_sw_loopback.yaml` pattern) without a physical NIC link.
- [ ] ResNet-50 ONNX → TensorRT engine conversion succeeds on first run; cached engine loads on subsequent runs.
- [ ] Feature vectors are extracted and PC1/PC2 values are printed to stdout.
- [ ] Image reassembly from reordered packets produces valid ResNet-50 input.
- [ ] Documentation updated per the docs-sync rules.
- [ ] Code passes `clang-format -style=file`.

## Hardware targets

- NVIDIA DGX Spark (GB10, sm_121)
- NVIDIA IGX Thor (H100/A100, sm_90/sm_80)
- x86_64 RTX Pro Server

## Dependencies

- TensorRT (≥ 8.6)
- ONNX model: ResNet-50 from the ONNX Model Zoo or `torch.onnx.export`
- CUDA runtime (already a project dependency)

chloecrozier:
Closes #73
## Changes
Adds the ResNet-50 example app: CIFAR-10 int8 (CIFAR-10 is native uint8) over raw Ethernet, engine-side GPU
reorder that converts int8→fp16 in the same pass, an SPSC ring between RX and
inference, TensorRT FP16 on the reorder output.

Cross-host on two DGX Sparks: 3,660 img/s, which is 4.4 Gb/s of image data
(4.7 Gb/s on the wire). That's GPU-bound, not network-bound — the same pair moves
~107 Gb/s into GPU memory with a light consumer, so one ResNet instance uses
about 4% of the ingest and there's headroom for a heavier model or more streams.

Also fixes an argv bug that made benches ignore `--seconds`.

## Test

```bash
cmake --build build -j --target daqiri_resnet50_inference   # -DDAQIRI_BUILD_APPLICATIONS=ON
applications/resnet50_inference/tools/run_resnet_xhost.sh --replay-once  # 256/256, dropped=0
mkdocs build --strict && python3 scripts/check_doc_refs.py

### CODE REVIEW

**Comment by RamyaGuru** [comment]
Thanks, @chloecrozier! A few suggestions for the benchmark results. I also want to test this on my Thor and I can take on some development tasks so we can knock this out before your trip. 

I think these should be quick to do. Just set it up and let it run.
1. Report all five model sizes: ResNet-18/34/50/101/152. I'm curious about ResNet-152, it was the only one that dropped packets before (~19%), so it's the cleanest test of whether the ring buffer absorbed that.
2. Longer benchmark runs: Looks like you currently run 8 inference batches for the benchmarks. I think we should do longer tests for better statistics. I did 3 reps of 30s each so I could put run-to-run error bars on it. Maybe do a quick time test (30s / 60s / 120s) at one model size first just to confirm the number is stable, then pick a duration and do 3 reps?
3. For latency, report the p50 and p99 statistics.

Some bigger updates:
1. Separate out DAQIRI and TensorRT throughput. Cliff had the idea of showing benchmarks like this to clearly show DAQIRI performance against TensorRT performance in this benchmark:

> - **DAQIRI alone** — simple RX test, just do reorder and int8 to fp16 and then free packets, measure the rate
> - **TensorRT alone** — no network at all, measure average inference kernel/graph completion time, repor images/s and Gbps
> - **Both** — what we're already measuring

2. From here, we should do a batch size sweep. I think we start with batch size 1 and increase in powers of 2 until throughput starts to dip. We should clear up how batch size is defined in the code. Are there separate batch size values specified for reorder and inference? Or are they fixed to be the same?

File: applications/resnet50_inference/configs/resnet50_rx_spark_xhost.yaml [hunk 1]
```diff
@@ -0,0 +1,114 @@
...
+bench_rx:
+- interface_name: "rx_port"
+  cpu_core: 19
+
+reorder:
+  reorder_name: "rx_reorder_resnet_int8_fp16"
+  out_payload_len: 1176
+  output_slot_stride: 2352
+  packets_per_image: 128
+  packets_per_batch: 4096
```
RamyaGuru: Maybe we should remove "packets_per_batch" as a config variable because it must be equal to packets_per_image x image_per_batch?
I also just noticed that there are two reorder-focused sections in the config file. Do we need both? Let's get some clarification on how inference batch size is handled in the code.

Issue #242, opened 2026-08-04
[BUG] IGX Thor DevKit raw GPUDirect Testing Follow-ups


RamyaGuru:
## Body

Four follow-ups from the IGX Thor raw GPUDirect session (aarch64, driver 580.00/CUDA 13.0, 256 MiB BAR1, ConnectX-7, daqiri `1134167`).

---

### 1. `build-container.sh` hardcodes CUDA 13.1 — unusable on IGX Thor (driver 13.0) [High]

`scripts/build-container.sh` unconditionally sets `DAQIRI_OS_BASE_IMAGE=nvcr.io/nvidia/cuda:13.1.0-devel-ubuntu24.04`. The IGX Thor BSP ships driver 580.00/CUDA 13.0; a 13.1 runtime cannot init against it (`cudaGetDeviceCount` → "unsupported display driver / cuda driver combination"; `cuda-compat` is datacenter-only and doesn't apply to Tegra). The Dockerfile already threads `DAQIRI_OS_BASE_IMAGE` as a build-arg but the wrapper script never exposes it.

**Fix:** let `DAQIRI_OS_BASE_IMAGE` (or a `CUDA_VERSION` shorthand) pass through from the environment instead of being hardcoded; document a `cuda:13.0.0-devel` override option. Optionally auto-detect the host driver's CUDA version.

Verified workaround: `--build-arg DAQIRI_OS_BASE_IMAGE=nvcr.io/nvidia/cuda:13.0.0-devel-ubuntu24.04` builds and runs fine (~1 min rebuild with DPDK layer cached).

---

### 2. NIC descriptor ring depth hardcoded at 8192 with no YAML config path [Medium-High]

`DpdkEngine::default_num_rx_desc` / `default_num_tx_desc` are fixed at 8192 with no way to override from YAML. `adjust_memory_regions` documents that a queue-backed MR needs `num_bufs ≥ ~3× ring` (= 24576) to avoid worker starvation (`rx_mbuf_allocation_errors` in the 100ks → flow-control pauses TX → ~885 pps). On a 256 MiB BAR1 GPU, two 8 KB-buffer GPUDirect regions cannot reach 24576 bufs (24576 × 8192 × 2 = 384 MiB), so the ring must shrink — but there is no supported way to do it.

[...]

PR #248, opened 2026-08-18
Issue 242 thor benchmark followups

RamyaGuru:
Addresses #242. All issues reproduced on current `main` (676b260) before being fixed.

**1. Size queue memory regions against `batch_size`, not just ring depth.**
`adjust_memory_regions()` sized MRs relative to the NIC ring only, ignoring that
`is_tx_burst_available()` needs `2x batch_size` free *on top of* the ring — so at the shipped
`batch_size: 10240` its own 3x-ring bump target (24576) sat below the 28672 TX requires and
silently wedged TX at `bursts=0`. Floor is now `max(1.5x ring, ring + 2x batch_size)` and the
bump target `max(3x ring, ring + 4x batch_size)`, plus a one-shot diagnostic for bursts the
pool can never satisfy.

**2. Fail fast when a device memory region has no GPUDirect path.**
With dma-buf unsupported and `nvidia_peermem` absent, registration logged "Successfully
registered" and then died two functions later as a bare `Invalid argument`; it now fails at
registration, naming the CUDA device and what else CUDA can see so "wrong GPU selected" is
readable off one line.

**3. Allow overriding the container CUDA base image.**
`scripts/build-container.sh` overwrote an exported `DAQIRI_OS_BASE_IMAGE`, pinning CUDA to
13.1.0 and making the script unusable on hosts with an older driver; it now honors a pre-set
value and adds `CUDA_VERSION` / `UBUNTU_VERSION`, with default output unchanged.

**Testing.** On IGX Thor with a ConnectX-7 cable loopback: bracketed the deadlock boundary
exactly, confirmed both previously-wedging configs now run at line rate with zero drops
(corroborated by `ethtool -S` wire counters), reproduced the GPUDirect failure on the Tegra
iGPU, and verified CUDA 13.0 inits where 13.1 returns `cuInit=803`. A/B against `main` with
matched build configuration shows no regressions across `gpudirect`, `hds`, the 4-queue
config, `sw_loopback`, and `reorder_seq`.

**Note:** #242's BAR1 premise is incorrect — the card has 8 GiB, not 256 MiB — so nothing
forces a smaller ring and the per-queue ring-depth YAML field the issue proposes is not
included; the sizing formula was the actual bug.


src/engines/dpdk/daqiri_dpdk_engine.cpp
```diff
@@ -1818,32 +1818,71 @@ void* DpdkEngine::alloc_huge(size_t bytes, int numa, AllocRegion::Deallocator* d
...
+    if (batch_it != mr_max_batch.end()) {
+      const uint32_t batch = batch_it->second;
+      const uint32_t min_num_bufs = std::max((ring_size * 3) / 2, ring_size + 2 * batch);
+      const uint32_t safe_num_bufs = std::max(ring_size * 3, ring_size + 4 * batch);
+      if (mr.second.num_bufs_ < min_num_bufs) {
+        DAQIRI_LOG_WARN(
+            "MR '{}' had num_bufs={}, below the {} required by a {}-deep NIC ring and "
+            "batch_size={} (ring + 2x batch); TX would wedge with no error and the RX refill "
+            "path would starve. Bumping to {} (ring + 4x batch), the smallest size measured "
+            "free of mbuf allocation errors.",
```
cliffburdick, line 1877: Should these warnings tell the user what to do? ie increase num_bufs?

src/engine_dpdk.cpp
```diff
@@ -45,6 +45,110 @@
...
+// real failure only surfaces later in map_memory_regions() as a bare rte_dev_dma_map "Invalid
+// argument" that names nothing (issue #242). Returns false when there is no usable path left.
+bool peermem_fallback_ok(const MemoryRegionConfig& mr, const char* why) {
+  if (nvidia_peermem_loaded()) {
+    return true;
+  }
+  DAQIRI_LOG_CRITICAL(
+      "Memory region '{}' requests kind: device on CUDA device {}, but {}, and the "
+      "nvidia_peermem module is not loaded -- there is no usable GPUDirect path for this "
+      "device. {}.",
```
cliffburdick, line 145: Suggest trying host_pinned?

cliffburdick (changes requested):
Does anything here apply to ibverbs path as well?

RamyaGuru:
> Does anything here apply to ibverbs path as well?

Yes, good call. I think at least item 2 on the list would also need to be ported over to ibverbs. I'll make sure I've gone through all the ibverbs testing and adjust the warning/helper messages based on your comments.

RamyaGuru:
@cliffburdick made updates to this PR to address your feedback and to make sure ibverbs is well-handled. Tested on both the IGX Thor DevKit (where warnings direct you to use the available discrete GPU with "device" MR) and GB10 (where warning direct you to use "host_pinned").

General implementation description:
This PR improves GPU selection on mixed-GPU systems such as IGX Thor by explicitly selecting the CUDA device specified by each memory region’s affinity before allocation and registration. It checks DMA-BUF support on the selected GPU and provides actionable guidance to choose a visible discrete GPU or use host_pinned memory when an integrated GPU cannot export DMA-BUF. It also preserves the caller’s CUDA context across these operations. Finally, it sizes raw DPDK buffers using both descriptor-ring and batch requirements and constrains ibverbs TX capacity to hardware limits, preventing stalls and opaque initialization failures.

cliffburdick (approved):

RamyaGuru:
Local validation run on commit 3e62553:
  - scripts/check_pr.sh
  - portable tests: 9 passed
  - mkdocs strict build: passed
  - HTML link check: passed
  - doc reference check: passed

PR #252, opened 2026-08-20
#250 - Surface 802.3x pause state and counters

### LINKED ISSUE #250: Surface 802.3x pause state and counters
I was digging into why the ResNet benchmark capped at 308 Gb/s and it turned out to
be 802.3x pause - 21.7% of 400 GbE line rate, with every drop counter sitting at
zero. Two gaps I hit: tune_system.py doesn't check pause, and pause counters never
show up in our xstats dump even though ethtool has nine of them.

chloecrozier:
Closes #250.

Found while chasing the ResNet benchmark's 308 Gb/s cap: 802.3x pause was eating 21.7% of
400 GbE line rate with every drop counter at zero, so it read as a slow transmitter. Two
things should have caught it and neither did.

**`tune_system.py` had no pause check.** Added `--check pause`: reports the pause parameters
per NIC, and separately whether the pause counters are non-zero, since pause that never fires
costs nothing. Also Step 10 in the system configuration guide and a warning in the raw
Ethernet benchmarking guide.

**Pause counters never reached the xstats dump.** They are not in `rte_eth_xstats_get` on
mlx5 at all, so the dump could not print them. `src/net_pause.{h,cpp}` reads them from the
kernel netdev instead (`ETHTOOL_GPAUSEPARAM` / `ETHTOOL_GSTATS`), and both raw engines call
it: warn at init when pause is enabled, print the counters with the stats dump. Counters are
per-run deltas against an init baseline, because the raw values are cumulative since boot and
would call a run throttled when pause was off for it.

## Test plan

- [x] 400 GbE loopback, pause on: warns at init, reports the frames exchanged during the run
- [x] No pause activity: reports `0`, no false "throttled" warning despite ~132M lifetime frames
- [x] Port with no netdev (software loopback): silent no-op
- [x] `--check pause` parsing: both-off, both-on, rx-only, tx-only, `RX/TX negotiated` lines
- [x] Build clean, `clang-format` clean, `check_doc_refs.py` and mkdocs anchors pass


docs/tutorials/system_configuration.md
```diff
@@ -1297,6 +1298,103 @@ DAQIRI requires an [**NVIDIA SmartNIC**](https://www.nvidia.com/en-us/networking
...
+
+    !!! warning "Leave pause enabled on lossless fabrics and against shallow-buffer peers"
+
+        On a lossless RoCE or PFC fabric, flow control is deliberate and required. Only disable pause on links dedicated to raw Ethernet where the application tolerates loss, and never on a shared fabric where your pause frames affect other tenants.
+
+        Pause frames on the wire are also not evidence of a misconfiguration by themselves. A peer that cannot buffer a line-rate burst — an FPGA is the common case, and it will pause often — asserts pause by design, and that is the mechanism working. Disabling pause on such a link does not recover the throughput; it converts the throttling into drops on the peer. The counters say which end asserted it: `rx_pause_ctrl_phy` counts frames **received** (the peer throttling this port's transmit) and `tx_pause_ctrl_phy` frames **sent** (this port's receive path throttling the peer). Non-zero `rx_pause_ctrl_phy` against a device known to be shallow-buffered is expected; non-zero `tx_pause_ctrl_phy` points back at this host.
```
cliffburdick: Consider adding something about using pacing rather than pause frames. Something like "If the receiver normally requires pause frames, packet pacing can instead be enabled to smooth out the packet rate to prevent buffer overruns".

docs/tutorials/system_configuration.md
```diff
@@ -1297,6 +1298,103 @@ DAQIRI requires an [**NVIDIA SmartNIC**](https://www.nvidia.com/en-us/networking
...
+    === "tune_system.py"
+
+        ```bash
+        sudo ./python/tune_system.py --check pause
+        ```
+
+        ??? abstract "See an example output"
+
+            ```
+            2026-08-19 09:43:00 - WARNING - Interface eth0 has 802.3x pause enabled (RX: on, TX: on). Link-level flow control can idle the link and cost over 20% of line rate with no drop counter to reveal it. Disable both directions with `ethtool -A eth0 rx off tx off` for raw-Ethernet benchmarking; keep it on lossless RoCE/PFC fabrics.
```
cliffburdick: I would remove the "Over 20%" since it's ambiguous and very system-dependent. Instead just say "can idle the link and prevent achieving higher rates"

cliffburdick (changes requested):
Couple small comments. Good PR overall

chloecrozier:
@cliffburdick thanks for reviewing, I just pushed the changes

PR #260, opened 2026-08-24
Register application-owned DPDK threads

tmartin-gh:
## Summary

- add engine-level registration hooks for application-owned dataplane threads
- register and unregister non-EAL application threads with DPDK
- keep no-op defaults for engines that do not require per-thread runtime state

## Motivation

Applications may construct or process packets on their own worker threads. DPDK APIs require those threads to be registered before using per-lcore state; doing this through the Engine interface avoids engine checks in shared application code.

## Testing

- exercised with application-owned TX construction workers in 22-queue DPDK runs

### DIFF
--- a/docs/api-reference/cpp.md
+++ b/docs/api-reference/cpp.md
@@ -653,8 +653,22 @@ workflow sections above show the common call order and ownership rules.
 | `parse_network_config(...)` | Parse YAML into `NetworkConfig` without starting the engine. |
 | `get_engine_type()` | Return the active engine type after initialization. |
 | `get_engine_type(config)` | Return the engine type selected by a config object. |
+| `register_current_thread()` | Initialize engine-specific state for the calling application-owned data-plane thread. |
+| `unregister_current_thread()` | Release state acquired by `register_current_thread()` on the calling thread. |
 | `shutdown()` | Stop DAQIRI and release engine-owned resources. |
 | `print_stats()` | Print engine statistics. |
+
+Application-created threads that call data-plane APIs should call
+`register_current_thread()` after initialization and before their first packet
+operation, then call `unregister_current_thread()` on that same thread before
+`shutdown()`. Registration is per thread and idempotent.
+
+For the DPDK engine, registration makes an application-owned thread known to
+the DPDK EAL. If the thread is already EAL-managed, registration succeeds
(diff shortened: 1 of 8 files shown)


cliffburdick:
I think the summary of this PR is not matching what the code does?

docs/api-reference/cpp.md
```diff
@@ -653,9 +653,23 @@ workflow sections above show the common call order and ownership rules.
 | `parse_network_config(...)` | Parse YAML into `NetworkConfig` without starting the engine. |
 | `get_engine_type()` | Return the active engine type after initialization. |
 | `get_engine_type(config)` | Return the engine type selected by a config object. |
+| `register_current_thread()` | Initialize engine-specific state for the calling application-owned data-plane thread. |
+| `unregister_current_thread()` | Release state acquired by `register_current_thread()` on the calling thread. |
 | `shutdown()` | Stop DAQIRI and release engine-owned resources. |
 | `print_stats()` | Print engine statistics. |
 
+Application-created threads that call data-plane APIs should call
```
cliffburdick, line 661: What's the reason this is needed? So far we've done all of our testing without EAL knowing about the other threads.

tmartin-gh: Good question. After tracing the current DPDK path, this is not required for correctness today. DPDK supports unregistered non-EAL pthreads using LCORE_ID_ANY, with library-specific limitations. The usual reason to register them is to enable per-lcore facilities—particularly the default mempool cache, which unregistered threads bypass.

However, all current DAQIRI DPDK packet, burst, and metadata pools are created with cache_size=0, so registration does not provide that performance benefit today either. DAQIRI’s (DPDK) NIC workers are already EAL-managed; application threads currently use ring and mempool/mbuf operations that work without registration. That explains why all existing testing has succeeded.

Registration could become useful if we enable per-lcore mempool caches, or necessary if we introduce a DPDK facility requiring a valid lcore ID. Until then, I think we could drop this change rather than impose an application lifecycle requirement without a demonstrated benefit. It also has a real tradeoff: rte_thread_register() is incompatible with DPDK multiprocess support (https://doc.dpdk.org/api-25.11/rte__lcore_8h.html)

tmartin-gh:
Decision to close.

PR #262, opened 2026-08-24
Clarify timestamp format

tmartin-gh:
## Summary

- define the public DPDK hardware timestamp contract as linear PTP epoch nanoseconds
- pass accurate-send timestamps directly to the DPDK timestamp dynamic field
- return driver-provided RX timestamps directly as nanoseconds
- perform no runtime clock-synchronization validation or calibration
- prominently document that timestamped I/O requires the NIC clock and `CLOCK_REALTIME` to be PTP-synchronized
- document the PTP clock-domain contract for DPDK and raw ibverbs

## Motivation

DAQIRI documents PTP-synchronized operation, but the DPDK engine left the unit and epoch of its public timestamp values implicit. A fixed linear-nanosecond contract lets callers compare received timestamps with `CLOCK_REALTIME` and schedule packets from that same clock without an engine-specific conversion.

## Compatibility

This is an intentional contract clarification: hardware timestamps are PTP epoch nanoseconds. DAQIRI does not validate that the NIC clock and `CLOCK_REALTIME` are synchronized. Timestamped operation without working PTP is unsupported and may produce incorrect scheduling and latency values. This requirement is documented prominently; DAQIRI performs no runtime validation or warning.

## Testing

- built both raw engines and all C++ examples in the CUDA 13.3 Aerial ARM container
- built the downstream dual-engine integration on ARM and x86
- validated full 22-queue bidirectional DPDK traffic with zero loss and zero timestamp failures on BlueField-3 integrated ConnectX-7 hardware

### DIFF
--- a/docs/api-reference/configuration.md
+++ b/docs/api-reference/configuration.md
@@ -347,10 +347,11 @@ a template table.
 
 `rx.hardware_timestamps:` Enable per-packet hardware RX timestamps for Raw Ethernet
 (`stream_type: "raw"`).
-When enabled, DAQIRI requires `RTE_ETH_RX_OFFLOAD_TIMESTAMP` support from the NIC/PMD and
-initialization fails if DAQIRI cannot provide nanosecond timestamps for the selected PMD.
-Timestamps are returned by `get_packet_rx_timestamp()` in nanoseconds in the NIC timestamp
-clock domain, not wall-clock time.
+When enabled, DAQIRI requires hardware timestamp support from the NIC and driver.
+Timestamps returned by `get_packet_rx_timestamp()` are unsigned 64-bit PTP epoch
+nanoseconds in the same clock domain as a PTP-synchronized `CLOCK_REALTIME`.
+**PTP synchronization is required.** DAQIRI does not validate the NIC or system clock
+configuration. Timestamp values are invalid if the clocks are not PTP-synchronized.
 
 - type: `boolean`
 - default: `false`
@@ -512,8 +513,12 @@ pre-encap (TX) frame.
 
 ### Accurate Send
 
-`tx.accurate_send:` Enable hardware-timed packet transmission using PTP timestamps. When
-enabled, use `set_packet_tx_time()` to schedule packets. Requires ConnectX-7 or later.
+`tx.accurate_send:` Enable hardware-timed packet transmission. When enabled, use
+`set_packet_tx_time()` to schedule packets. The supplied timestamp is always an unsigned
+64-bit PTP epoch-nanosecond value in the same clock domain as a PTP-synchronized
+`CLOCK_REALTIME`. **PTP synchronization is required.** DAQIRI does not validate the NIC or
+system clock configuration. Scheduled transmission is invalid if the clocks are not
+PTP-synchronized. Requires ConnectX-7 or later.
 
 - type: `boolean`
 - default: `false`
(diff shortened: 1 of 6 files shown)


include/daqiri/common.h
```diff
@@ -1423,6 +1448,27 @@ template <> struct YAML::convert<daqiri::NetworkConfig> {
...
+              } else if (format == "nanoseconds") {
+                tx_cfg.hardware_timestamp_format_ =
+                    daqiri::TxTimestampFormat::NANOSECONDS;
+              } else {
+                DAQIRI_LOG_ERROR(
+                    "Invalid tx.hardware_timestamp_format '{}' for interface '{}': "
+                    "expected 'device_clock_ticks' or 'nanoseconds'",
+                    format, ifcfg.name_);
+                return false;
+              }
```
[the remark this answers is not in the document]
tmartin-gh: Please re-review

tmartin-gh:
Greptile please update the summary

docs/api-reference/configuration.md
```diff
@@ -333,8 +333,14 @@ initialization fails if DAQIRI cannot provide nanosecond timestamps for the sele
 Timestamps are returned by `get_packet_rx_timestamp()` in nanoseconds in the NIC timestamp
 clock domain, not wall-clock time.
 
+`rx.hardware_timestamp_format:` Describes the hardware/PMD representation DAQIRI receives.
+Use `device_clock_ticks` (default) when DAQIRI must convert device ticks to nanoseconds, or
+`nanoseconds` when the PMD already supplies nanoseconds.
```
cliffburdick: PMD is specific to DPDK, so maybe this should just say "driver"

docs/api-reference/configuration.md
```diff
@@ -496,8 +502,22 @@ pre-encap (TX) frame.
...
+`tx.hardware_timestamp_format:` Selects the clock-domain representation accepted by
+`set_packet_tx_time()`:
+
+- `device_clock_ticks` preserves the historical contract and passes device-clock values through.
+- `nanoseconds` accepts epoch nanoseconds. The DPDK engine validates at startup that
+  `rte_eth_read_clock()` is readable and agrees with `CLOCK_REALTIME`; initialization fails
+  rather than sending at an incorrectly interpreted time when mlx5 real-time clock mode or PTP
+  synchronization is unavailable.
+
+The raw-ibverbs engine rejects `nanoseconds` until wall-clock send-CQ/WAIT encoding support is
```
cliffburdick: Since this is a user document let's remove CQ/WAIT terminology and just put something like "until wall-clock encoding support"...

cliffburdick (commented):
Just a general comment, but shouldn't we always expect nanoseconds? In what case would a user want device ticks?

tmartin-gh:
Greptile:

This feedback assumes that the DPDK backend supports selectable device-tick and epoch-nanosecond clock domains. It does not. The DPDK timestamp contract is intentionally epoch nanoseconds only.
RX returns the timestamp representation supplied by the supported driver configuration without applying a device-counter calibration. TX accepts timestamps in that same epoch-nanosecond representation and intentionally passes them directly to the driver. Consequently, there is no alternate DPDK timestamp format to convert, preserve, or select at either boundary.
We also intentionally do not validate PTP synchronization in DAQIRI. Configuring and synchronizing the NIC clock is a deployment prerequisite. An unsynchronized clock can produce incorrect scheduling or latency results, but supporting or detecting that operational condition is outside this API’s scope.

Issue #257, opened 2026-08-21
[FEA] Allow externally-supplied user buffers


cliffburdick:
We should support packet buffers provided by the user. This would let users allocate buffers in different cuda contexts than our startup context.

PR #263, opened 2026-08-24
#257 - Add external memory region bindings

cliffburdick:


dleshchev:
Validation gap from the benchmark pass: the existing loopback runs still only exercise DAQIRI-owned memory. They build and run cleanly, but they do not call `get_memory_region_requirements(...)`, pass `MemoryRegionBindings` into `daqiri_init(...)`, use `owned=false` regions, or cover the caller-owned CUDA context / capacity / alignment paths added here.

Since this PR adds a new public C++/Python API plus engine registration behavior for external regions, I think it needs a minimal direct smoke path before merge. A small C++ example or benchmark flag would be enough: parse the software-loopback config, query requirements, `cudaMalloc` caller-owned TX/RX buffers, initialize with bindings, run briefly, shutdown, then free the buffers. Without that, this new API could regress while the current benchmark suite still passes.

dleshchev (commented):
see comment above, otherwise good to merge

docs/api-reference/cpp.md
```diff
@@ -56,6 +56,44 @@ Only one engine may be active in a process. Calling `daqiri_init()` again before
...
+
+// Return every outstanding burst before shutdown.
+daqiri::shutdown();
+cudaFree(rx_gpu);
+```
+
+Bindings may cover only some regions; DAQIRI allocates the rest. Bound memory is borrowed, so its
+allocation and CUDA context must remain alive until `shutdown()` completes. DAQIRI deregisters it
+from the NIC but never frees, unpins, or unmaps it. Direct TCP/UDP sockets reject bindings because
+their configured regions are not packet pools. DPDK also rejects externally bound `huge` regions:
```
tmartin-gh, line 93: We have successfully used a combination of rte_extmem_register(), rte_dev_dma_map(), and mlock() to use externally allocated memory with DPDK.

tmartin-gh:
General comment: Tested this PR out locally on my setup with an MPS application, it was functional and performant.  No other concerns at this time.

Issue #255, opened 2026-08-21
[FEA] Add hardware loopback


cliffburdick:
Software loopback exists in daqiri with a limited set of functionality. Add a hardware loopback mode that removes all the restrictions of software loopback, and works with all the normal non-loopback features (header-data split, flow steering, etc).

PR #264, opened 2026-08-24
#255 - Add ibverbs hardware loopback

cliffburdick:
We have software loopback for sending/receiving packets on systems without a physical cable. However, SW loopback has many problems, including no flow steering, single queues, and poor performance. If available, hardware flow steering is better in all aspects, and should be used instead. 

This PR introduces HW flow steering by leveraging a QP flag to allow unicast traffic to loop back on the same port. Only available for ibverbs backend at the moment.


dleshchev:
Docs clarification from validation: this benchmark should still follow the `mlnx_perf` guidance, but because `loopback: "hw"` uses NIC-internal loopback, the physical SerDes counters are expected to stay near zero. Users should measure this config from `mlnx_perf`'s vport/internal-loopback counters, especially `vport_loopback_bytes` / `vport_loopback_packets` (with `rx_vport_unicast_bytes` tracking similarly), rather than treating `tx_bytes_phy` / `rx_bytes_phy` as the throughput source.

For real on-wire validation, the docs should continue to direct users to a physical or cross-host setup and the `*_phy` counters. I think adding this note near the new hardware-loopback config or the existing `mlnx_perf` section would avoid confusion.

cliffburdick:
> Docs clarification from validation: this benchmark should still follow the `mlnx_perf` guidance, but because `loopback: "hw"` uses NIC-internal loopback, the physical SerDes counters are expected to stay near zero. Users should measure this config from `mlnx_perf`'s vport/internal-loopback counters, especially `vport_loopback_bytes` / `vport_loopback_packets` (with `rx_vport_unicast_bytes` tracking similarly), rather than treating `tx_bytes_phy` / `rx_bytes_phy` as the throughput source.
> 
> For real on-wire validation, the docs should continue to direct users to a physical or cross-host setup and the `*_phy` counters. I think adding this note near the new hardware-loopback config or the existing `mlnx_perf` section would avoid confusion.

Thanks. I'll update it.

dleshchev (approved):

PR #265, opened 2026-08-24
Decode raw-ibverbs real-time RX timestamps

tmartin-gh:
## Summary

- configure mlx5 RQs for real-time CQEs when nanoseconds are requested
- decode mlx5 UTC CQE timestamps into epoch nanoseconds
- preserve free-running/device-clock behavior by default
- document the raw-ibverbs timestamp contract

## Motivation

Returning the raw CQE field as nanoseconds is incorrect for real-time mlx5 CQEs because the hardware stores seconds and nanoseconds in separate bit fields. The RQ timestamp format and software conversion must agree.

## Compatibility

The existing device-clock path is unchanged. Conversion is enabled only for the explicit `nanoseconds` mode.

## Testing

- validated per-packet epoch RX timestamps and latency histograms in 22-queue raw-ibverbs runs

### DIFF
--- a/include/daqiri/common.h
+++ b/include/daqiri/common.h
@@ -1337,6 +1337,31 @@ template <> struct YAML::convert<daqiri::NetworkConfig> {
             } catch (const std::exception& e) { rx_cfg.flow_isolation_ = false; }
 
             try {
+              rx_cfg.hardware_timestamps_ = rx["hardware_timestamps"].as<bool>();
+            } catch (const std::exception& e) { rx_cfg.hardware_timestamps_ = false; }
+
+            try {
+              const auto format =
+                  rx["hardware_timestamp_format"].as<std::string>();
+              if (format == "device_clock_ticks") {
+                rx_cfg.hardware_timestamp_format_ =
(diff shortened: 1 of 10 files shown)


include/daqiri/common.h
```diff
@@ -1336,6 +1336,31 @@ template <> struct YAML::convert<daqiri::NetworkConfig> {
               rx_cfg.flow_isolation_ = rx["flow_isolation"].as<bool>();
             } catch (const std::exception& e) { rx_cfg.flow_isolation_ = false; }
 
+            try {
+              rx_cfg.hardware_timestamps_ = rx["hardware_timestamps"].as<bool>();
+            } catch (const std::exception& e) { rx_cfg.hardware_timestamps_ = false; }
+
+            try {
+              const auto format =
+                  rx["hardware_timestamp_format"].as<std::string>();
```
cliffburdick, line 1345: This seems like it has duplicate code with #262 ?

tmartin-gh:
Closing, will open consolidated ibverbs improvements PR.

PR #266, opened 2026-08-24
Group TX queues by worker core

tmartin-gh:
## Summary

- group TX queues that map to the same worker core
- launch each unique DPDK lcore once
- have one worker service all queues assigned to that core
- size packet-pool availability for queued and in-flight work
- preserve pacing state across partial submissions, free every unchained scatter segment during shutdown, and reject RX/TX worker overlap on one lcore

## Motivation

DAQIRI's explicit queue-to-core mapping may repeat a core for multiple queues. Treating the mapping length as the worker count attempts to launch the same EAL lcore repeatedly and leaves queues without a valid worker.

## Testing

- built both raw engines and all C++ examples in the CUDA 13.3 Aerial ARM container
- validated mappings of 22 TX queues across two worker cores and, separately, onto one worker core

### DIFF
--- a/src/engines/dpdk/daqiri_dpdk_engine.cpp
+++ b/src/engines/dpdk/daqiri_dpdk_engine.cpp
@@ -454,6 +454,10 @@ struct TxWorkerParams {
   uint64_t pace_rem = 0;
 };
 
+struct TxWorkerMultiQParams {
+  std::vector<TxWorkerParams> q_params;
+};
+
 struct RxWorkerParams {
   int port;
   int queue;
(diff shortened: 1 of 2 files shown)


src/engines/dpdk/daqiri_dpdk_engine.cpp
```diff
@@ -5851,8 +5901,20 @@ void DpdkEngine::run() {
...
+        const int tx_core = strtol(q.common_.cpu_core_.c_str(), NULL, 10);
+        TxWorkerParams *params;
+        if (loopback_ == LoopbackType::LOOPBACK_TYPE_SW) {
+          params = new TxWorkerParams{};
+        } else {
+          auto &core_params = tx_core_params[tx_core];
+          if (core_params == nullptr) {
+            core_params = new TxWorkerMultiQParams;
+          }
+          core_params->q_params.emplace_back();
```
cliffburdick, line 5966: Is this using a single ring from all senders? Is it MPSC?

cliffburdick (commented):
One outstanding comment

src/engines/dpdk/daqiri_dpdk_engine.cpp
tmartin-gh: Legacy DPDK engine code uses MP/MC.  I didn't change it.  Legacy documentation https://github.com/NVIDIA/daqiri/blob/main/docs/concepts.md#L241-L246 indicates "Queues and cores are not strictly one-to-one: one core can service multiple queues."

cliffburdick (approved):

PR #267, opened 2026-08-24
Preserve CUDA memory-region context ownership

tmartin-gh:
## Summary

- record the CUDA context that owns each device memory region
- allocate in an existing application/MPS context instead of unconditionally switching to the primary context
- reactivate the owning context while freeing device memory, then restore the caller's context
- avoid changing CUDA contexts while raw ibverbs registers GPUDirect memory

## Motivation

Calling `cudaSetDevice()` during allocation, registration, or destruction can replace an application's current MPS subcontext with the device primary context. CUDA device pointers are context-owned, so later GPUDirect registration or cleanup can run against the wrong context and fail or corrupt resource lifetime.

## Compatibility

Callers without a current CUDA context retain primary-context initialization. For every configured region, DAQIRI selects the owning GPU context only for allocation and restores the context that was current on entry, including when consecutive regions target different GPUs.

## Testing

- built `daqiri_common` and `daqiri_dpdk` standalone in the CUDA 13.3 Aerial x86 container
- exercised the combined change in validated 22-queue DPDK and raw-ibverbs hardware runs using separate MPS contexts

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -191,12 +191,27 @@ void Engine::free_memory_regions() noexcept {
         }
         cudaFreeHost(region.ptr_);
         break;
-      case AllocRegion::Deallocator::CUDA_DEVICE:
-        if (region.affinity_ >= 0) {
-          cudaSetDevice(region.affinity_);
+      case AllocRegion::Deallocator::CUDA_DEVICE: {
+        CUcontext previous = nullptr;
+        const CUresult get_result = cuCtxGetCurrent(&previous);
+        if (get_result != CUDA_SUCCESS) {
+          DAQIRI_LOG_ERROR("Could not query current CUDA context while freeing MR {}", name);
+          break;
         }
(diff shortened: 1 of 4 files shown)


src/engine.cpp
```diff
@@ -447,16 +462,60 @@ Status Engine::allocate_memory_regions() {
...
+          CUcontext current = nullptr;
 
-          cudaSetDevice(mr.second.affinity_);
-          cudaFree(0);  // Create primary context if it doesn't exist
+          const auto current_res = cuCtxGetCurrent(&current);
+          if (current_res != CUDA_SUCCESS) {
+            DAQIRI_LOG_CRITICAL("Could not query the current CUDA context");
+            return Status::NULL_PTR;
+          }
+          const CUcontext previous = current;
```
cliffburdick, line 472: current or current_res?

cliffburdick (approved):

PR #268, opened 2026-08-24
Drain TX before releasing engine resources

tmartin-gh:
## Summary

- add an Engine API for waiting until submitted TX work is idle
- implement drain checks for DPDK and raw ibverbs
- stop and join workers before destroying flows, ports, rings, EAL state, and registered memory
- propagate TX-idle timeout failures instead of silently continuing teardown

## Motivation

The previous shutdown sequence could free rings, ports, or memory registrations while TX workers or the NIC still referenced them. This caused intermittent shutdown crashes and resource-lifetime violations.

## Testing

- validated clean startup/shutdown after full 22-queue DPDK and raw-ibverbs runs

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -935,6 +935,11 @@ Status Engine::get_tx_packet_burst_checked(BurstParams* burst) {
   return get_tx_packet_burst(burst);
 }
 
+Status Engine::wait_for_tx_idle(uint32_t timeout_ms) {
+  (void)timeout_ms;
+  return Status::NOT_SUPPORTED;
+}
+
 Status Engine::get_rx_burst(BurstParams** burst, int port_id) {
   // Check if the port_id is valid
   if (port_id < 0 || port_id >= static_cast<int>(cfg_.ifs_.size())) {
--- a/src/engine.h
+++ b/src/engine.h
@@ -110,6 +110,7 @@ class Engine {
   virtual void free_tx_metadata(BurstParams* burst) = 0;
   virtual Status get_tx_metadata_buffer(BurstParams** burst) = 0;
   virtual Status send_tx_burst(BurstParams* burst) = 0;
+  virtual Status wait_for_tx_idle(uint32_t timeout_ms);
   virtual Status get_mac_addr(int port, char* mac) = 0;
   virtual Status drop_all_traffic(int port);
   virtual Status allow_all_traffic(int port);
--- a/src/engine.h
+++ b/src/engine.h
@@ -108,7 +108,6 @@ class Engine {
   virtual Status get_reorder_burst_info(BurstParams* burst, ReorderBurstInfo* info);
   virtual void free_rx_metadata(BurstParams* burst) = 0;
   virtual void free_tx_metadata(BurstParams* burst) = 0;
-  virtual Status get_tx_metadata_buffer(BurstParams** burst) = 0;
   virtual Status send_tx_burst(BurstParams* burst) = 0;
   virtual Status wait_for_tx_idle(uint32_t timeout_ms);
   virtual Status get_mac_addr(int port, char* mac) = 0;
--- a/src/engines/dpdk/daqiri_dpdk_engine.cpp
+++ b/src/engines/dpdk/daqiri_dpdk_engine.cpp
@@ -6905,18 +6905,152 @@ void DpdkEngine::free_tx_metadata(BurstParams* burst) {
   rte_mempool_put(tx_metadata, burst);
 }
 
-Status DpdkEngine::get_tx_metadata_buffer(BurstParams** burst) {
-  if (rte_mempool_get(tx_metadata, reinterpret_cast<void**>(burst)) != 0) {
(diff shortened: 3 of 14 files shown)


src/engines/dpdk/daqiri_dpdk_engine.cpp
```diff
@@ -6852,12 +6852,20 @@ void DpdkEngine::free_tx_metadata(BurstParams* burst) {
...
 }
 
-Status DpdkEngine::get_tx_metadata_buffer(BurstParams** burst) {
-  if (rte_mempool_get(tx_metadata, reinterpret_cast<void**>(burst)) != 0) {
-    DAQIRI_LOG_CRITICAL("Running out of TX meta buffers due to high rates. Either increase "\
-      "your number of metadata buffers (current: {}) with `tx_meta_buffers` (will "\
-      "increase memory usage) or increase your `batch_size` for port {} queue {} (will increase "\
-      "latency)", cfg_.tx_meta_buffers_, (*burst)->hdr.hdr.port_id, (*burst)->hdr.hdr.q_id);
+Status DpdkEngine::get_tx_metadata_buffer(BurstParams **burst) {
+  if (burst == nullptr) {
```
cliffburdick: These extra checks and assignments seem to be the same as the user just checking for an error code, or am I reading that wrong?

src/engines/dpdk/daqiri_dpdk_engine.cpp
```diff
@@ -6866,6 +6874,131 @@ Status DpdkEngine::get_tx_metadata_buffer(BurstParams** burst) {
...
+  };
+
+  const auto deadline = steady_clock::now() + milliseconds(timeout_ms);
+  do {
+    if (idle()) {
+      return Status::SUCCESS;
+    }
+    std::this_thread::sleep_for(std::chrono::microseconds(50));
+  } while (steady_clock::now() < deadline);
+  if (tx_metadata != nullptr && !rte_mempool_full(tx_metadata)) {
```
cliffburdick, line 6991: newline

src/engine.cpp
```diff
@@ -191,12 +191,27 @@ void Engine::free_memory_regions() noexcept {
         }
         cudaFreeHost(region.ptr_);
         break;
-      case AllocRegion::Deallocator::CUDA_DEVICE:
-        if (region.affinity_ >= 0) {
-          cudaSetDevice(region.affinity_);
+      case AllocRegion::Deallocator::CUDA_DEVICE: {
+        CUcontext previous = nullptr;
+        const CUresult get_result = cuCtxGetCurrent(&previous);
```
cliffburdick, line 196: Code from a different PR

src/engines/dpdk/daqiri_dpdk_engine.cpp
```diff
@@ -6866,6 +6874,131 @@ Status DpdkEngine::get_tx_metadata_buffer(BurstParams** burst) {
   return Status::SUCCESS;
 }
 
+Status DpdkEngine::wait_for_tx_idle(uint32_t timeout_ms) {
```
cliffburdick, line 6908: Where would this function be called?

src/engine.cpp
tmartin-gh: Will be removed

src/engines/dpdk/daqiri_dpdk_engine.cpp
tmartin-gh: A couple things:

1. The DPDK version is currently de-referencing *burst after an rte_mempool_get() fails.  This is likely going to result in a bad pointer dereference.  The code fixes this.

2. I added a 'burst' nullptr check as defensive.  It's not necessary, and in fact the ibverbs doesn't have it.  Since this is a virtual function, there's no higher level dispatch, so if we wanted defensive programming we'd keep the one here and put one in the ibverbs version too.

3. The *burst = nullptr assignment is defensive.  Looking at the ibverbs get_tx_metadata_buffer implementation, it's clear *burst is nullptr on a failure, so making the nullptr assignment here just gives parity between the two implementations.  We could move the *burst = nullptr into the failure path instead of always doing it.
```
Status IbverbsEngine::get_tx_metadata_buffer(BurstParams** burst) {
  *burst = create_tx_burst_params();
  return *burst ? Status::SUCCESS : Status::NO_FREE_BURST_BUFFERS;
}
```

tmartin-gh: get_tx_metadata_buffer() was obsolete/dead code.  Removed it.

src/engine.h
```diff
@@ -46,6 +47,7 @@ struct AllocRegion {
   void* ptr_ = nullptr;
   size_t size_ = 0;
   int affinity_ = -1;
+  CUcontext cuda_context_ = nullptr;
```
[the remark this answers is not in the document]
tmartin-gh: Did this get addressed to your satisfaction?

cliffburdick (approved):

PR #271, opened 2026-08-24
Reuse raw-ibverbs flow matchers

tmartin-gh:
## Summary

- cache raw-ibverbs DR matchers by criteria and mask
- reuse a matcher for rules with the same match shape
- retain independent per-flow rules and destinations
- document matcher cache identity and per-rule ownership

## Motivation

Creating one mlx5 DR matcher per flow exhausts or stresses steering resources as queue count grows. Multiple MAC-differentiated queues use the same mask and can safely share a matcher.

## Testing

- validated 22 MAC-steered raw-ibverbs RX queues with zero packet loss

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -419,6 +419,14 @@ for the configuration constraints and
 [C++ API Usage → Reordered RX bursts](api-reference/cpp.md#reordered-rx-bursts)
 for how to consume them from C++.
 
+### Raw ibverbs flow matcher reuse
+
+The raw ibverbs engine caches matchers by their complete criteria-enable value
+and match mask. Flow rules that share that matcher key reuse the matcher while
+retaining independent match values, actions, destinations, and rule lifetimes.
+Rules with different criteria or masks never share a matcher.
+
+
 ## See also
 
 - [API Guide](api-reference/index.md): the 6-step DAQIRI application
--- a/src/engines/ibverbs/daqiri_ibverbs_engine.cpp
+++ b/src/engines/ibverbs/daqiri_ibverbs_engine.cpp
@@ -1868,19 +1868,41 @@ Status IbverbsEngine::install_port_flows() {
     // Build the per-flow rules. Each flow with a queue action steers its matched
     // 5-tuple to that queue's TIR. A flow that matches no supported field, and
     // the no-flows case, fall through to a catch-all -> the first queue.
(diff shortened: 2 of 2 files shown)


cliffburdick (commented):
It would be good to understand the use of this and what problem it's trying to solve in the current library. Is flow match creation too slow?

tmartin-gh:
> It would be good to understand the use of this and what problem it's trying to solve in the current library. Is flow match creation too slow?

This is an optimization to reduce calls to `mlx5dv_dr_matcher_create()` and whatever resources are consumed by creation of matchers.  

Imagine a scenario where we want to do flow steering based on source MAC address, we 22 such different source MAC addresses, and 2 rx queues.  With the current DAQIRI code we have something like:
```
  matcher(src-MAC mask) + rule(src MAC 0 → RXQ 0)
  matcher(src-MAC mask) + rule(src MAC 1 → RXQ 1)
...
  matcher(src-MAC mask) + rule(src MAC 20 → RXQ 0)
  matcher(src-MAC mask) + rule(src MAC 21 → RXQ 1)
```

and would create 22 'matchers' with the exact same specifications.  The matcher represents a hardware steering mask/template and can own or cause allocation of steering structures/device memory as rules are added. The rdma-core API explicitly says one matcher is designed to hold multiple rules with an identical mask, and exposes matcher layout/resource sizing plus device-memory reclamation.

This PR implements the optimization where one matcher can be combined with multiple rules.  For the same scenario described above we would have something like:
```
one matcher(src-MAC mask)
  ├── src MAC 0  → RXQ 0
  ├── src MAC 1  → RXQ 1
  ├── ...
  └── src MAC 21 → RXQ 1
```

However, we can defer the PR, it's not critical for performance nor is the example above a burden on resource consumption at its current scale.

docs/concepts.md
```diff
@@ -419,6 +419,14 @@ for the configuration constraints and
 [C++ API Usage → Reordered RX bursts](api-reference/cpp.md#reordered-rx-bursts)
 for how to consume them from C++.
 
+### Raw ibverbs flow matcher reuse
```
cliffburdick, line 422: This shouldn't need to be documented. It's an implementation detail.

cliffburdick (changes requested):

tmartin-gh:
Decision to defer/not merge for now. Closing.

PR #273, opened 2026-08-24
Flush raw-ibverbs BlueFlame writes

tmartin-gh:
## Summary

- add an architecture-specific write-combining MMIO flush helper
- order the mlx5 doorbell record and BlueFlame write
- flush the BlueFlame write before reusing the alternating BF offset
- document the write-combining ordering requirement

## Motivation

A DMA/store barrier orders descriptors and the doorbell record but does not guarantee that a write-combining MMIO store has reached the device. This is observable on Arm when the BlueFlame offset is reused quickly.

## Testing

- exercised on GH200 during sustained 22-queue raw-ibverbs accurate-TX runs


tmartin-gh:

PR #274, opened 2026-08-24
Do not offload checksums for raw ibverbs

tmartin-gh:
## Summary

- stop requesting IPv4 and L4 checksum offloads on raw-ibverbs SEND WQEs
- apply the correction to ordinary SEND and eMPW paths
- document that DAQIRI transmits complete raw Ethernet frames without assuming an upper-layer protocol
- compute IPv4 header checksums in the shipped raw IPv4 examples

## Motivation

The raw stream API accepts arbitrary EtherTypes and complete application-provided frames. Unconditionally requesting IP/L4 checksum offload can make the NIC reinterpret and modify non-IP payloads, including eCPRI-style benchmark traffic.

## Compatibility

Applications remain responsible for constructing the complete raw Ethernet frame, including any checksums required by their selected protocol.

## Testing

- built the `daqiri_ibverbs` target in the CUDA 13.3 Aerial ARM container
- validated the combined implementation with EtherType `0xaefe` in full 22-queue raw-ibverbs hardware runs with zero packet loss


tmartin-gh:

PR #275, opened 2026-08-24
Schedule timed raw-ibverbs bursts with eMPW

tmartin-gh:
## Summary

- allow enhanced multi-packet WQEs for bursts timed only on packet zero
- emit one WAIT and ordinary SEND for the timed first packet
- use `ONLY_FIRST_ERR` completion mode for the new unsignaled WAIT and SEND WQEs
- pack the remaining untimed packets into eMPW WQEs
- account for the exact SQ space required by the mixed WQE sequence
- request IPv4/L4 checksum offload on the ordinary first-packet SEND, matching the existing raw SEND contract on `main`

## Motivation

Accurate burst scheduling applies one release timestamp to the first packet, while the remaining packets must follow immediately. Falling back to one ordinary SEND per packet consumes excessive SQ capacity at high queue counts. A WAIT plus first SEND followed by eMPW preserves the scheduling contract while substantially reducing WQEBB use.

## Compatibility

Untimed eMPW behavior is retained. Bursts with multiple per-packet timestamps or unsupported segment layouts continue to use the ordinary SEND path.

## Testing

- built the complete project with the DPDK and raw-ibverbs engines in the CUDA 13.3 Aerial ARM container
- validated the combined implementation in full 22-queue raw-ibverbs hardware runs with zero packet loss


tmartin-gh:

PR #281, opened 2026-08-28
ibverbs improvements

tmartin-gh:
## Summary

- use PTP epoch nanoseconds for raw-ibverbs timed transmission and RX hardware timestamps
- reduce the striding-RQ packet region to the capacity required by the configured queues
- register the DevX RX work queue with the correct UMEM access and ownership
- use the required TX completion mode for unsignaled raw-ibverbs WQEs
- order and flush write-combining BlueFlame doorbell writes before reusing the BF offset
- schedule a timed first packet with WAIT/SEND WQEs and pack the remaining burst with eMPW
- coalesce consecutive striding-RQ packet releases by WQE in the bulk RX free path

## Motivation

A 22 flow near-200Gbps-line-rate bidirectional pattern runs EndpointA TX, EndpointA RX, EndpointB TX, and EndpointB RX as four separate applications. At that queue count, the previous raw-ibverbs paths consumed excessive receive memory and TX SQ capacity, performed unnecessary atomic RX-release updates, and could expose DevX registration, completion, and BlueFlame ordering defects. These changes are the functional and performance fixes required to sustain the pattern on GH200 GPU memory, GH200 CPU memory, and x86 CPU memory.

This PR consolidates the retained work from #269 and #272 through #275, plus the raw-ibverbs PTP functionality retained from #265 and #270. It intentionally excludes the matcher-reuse work from closed PR #271 and does not support device-tick timestamps.

## Compatibility

- raw-ibverbs timestamps use unsigned 64-bit PTP epoch nanoseconds; DAQIRI does not validate PTP synchronization
- untimed eMPW and individual-packet RX release behavior remain available
- no DPDK behavior is changed by this PR

## Testing

- completed clean Release builds with DPDK and raw-ibverbs enabled on GH200/aarch64 and x86_64
- passed the application unit-test suite on both architectures, with only the expected GPU-only test skipped
- validated four simultaneous EndpointA TX, EndpointA RX, EndpointB TX, and EndpointB RX applications at the full 22-queue pattern
- validated both GH200 GPU/device memory and GH200 CPU/hugepage memory with x86 CPU/hugepage memory
- received every measured packet with zero loss, duplicates, reorder, missteering, invalid or foreign packets, or timestamp failures
- completed every TX direction with zero dropped or skipped bursts and zero late build or enqueue events

PR #284, opened 2026-08-31
#283 - Gate ibverbs real-time RX timestamps

### LINKED ISSUE #283: [BUG] ibverbs raw RX fails CREATE_RQ when real-time CQ timestamps are unsupported
**Describe the bug**

The ibverbs raw-RX benchmark on `main` fails during `daqiri_init()` while creating the DevX receive queue. The same raw-RX configuration initializes successfully with an image built before #281.

The failure is an initialization regression, not a throughput result:

```text
[CRITICAL] CREATE_RQ (DevX) failed: Remote I/O error (syndrome 0x43d98f)
```

#281 made the RQ request `MLX5_RQC_TIMESTAMP_FORMAT_REAL_TIME` unconditionally. The affected device does not advertise `MLX5DV_CONTEXT_FLAGS_REAL_TIME_TS`, and its firmware rejects that unsupported RQ timestamp format.

Individually restoring the other RX-side changes did not resolve the failure:

- Restoring the previous striding-RQ geometry still fails with syndrome `0x43d98f`.
- Restoring the previous RX WQ UMEM access flags still fails with syndrome `0x43d98f`.
- Restoring both still fails.
- Omitting only the unsupported real-time timestamp request allows the identical configuration to initialize.

1. Build current `main` in the project container with:

   ```bash
   BASE_TARGET=dpdk DAQIRI_ENGINE="dpdk ibverbs" scripts/build-container.sh
   ```

2. On the affected ARM64 system, run `daqiri_bench_raw_gpudirect` in RX-only mode with a valid single-port raw ibverbs configuration using a GPU memory region and one striding RX queue.
3. Observe that `daqiri_init()` fails at DevX `CREATE_RQ` before traffic or throughput measurement begins.
4. Run the same configuration with a pre-#281 image and observe successful initialization.

No machine-local interface names, PCI addresses, MAC addresses, network addresses, or hostnames are required to reproduce the issue; substitute locally valid values in the example raw-RX configuration.

**Expected behavior**

Raw ibverbs RX initialization should create the DevX RQ successfully on devices that do not support real-time CQ timestamps. DAQIRI should request the real-time format only when `mlx5dv_query_device()` advertises `MLX5DV_CONTEXT_FLAGS_REAL_TIME_TS`; otherwise it should retain the device-clock CQ format and convert timestamps through the existing clock-conversion path.

The valid #281 changes should remain intact: RX WQ UMEM registration with no remote-access permissions, the reduced 64-stride region geometry, and the adapter-relative `log_wq_pg_sz` encoding.

- Environment location: bare-metal host; DAQIRI runs in the privileged project container
- Method of DAQIRI install: project container built from source
- Container build: `BASE_TARGET=dpdk DAQIRI_ENGINE="dpdk ibverbs" scripts/build-container.sh`

**Environment details**

- OS: Ubuntu 24.04.4 LTS
- Architecture/platform: ARM64 NVIDIA DGX Spark
- Kernel: `6.17.0-1014-nvidia`
- Kernel page size currently measured on both validation systems: 4096 bytes
- NIC: NVIDIA ConnectX-7, MT2910 family (`15b3:1021`)
- NIC firmware: `28.45.4028`
- rdma-core, libibverbs, and providers: `2407mlnx52-1.2407061`
- DAQIRI: `main` containing `dbf0c8b` (regression); pre-#281 image succeeds

**Additional context**

The mlx5 WQ page-size field is encoded relative to the adapter's fixed 4-KiB page. It is distinct from the page size used to align/register the UMEM and was not the cause in the individual-revert matrix. Increasing the buffer count to restore the previous effective queue depth also does not fix initialization.

A narrow fix will capability-gate the RQ timestamp format and select the matching timestamp decode path per queue. Hardware reproduction is required because this failure occurs in firmware during DevX object creation.

dleshchev:
Fixes #283.

## Summary

- Request mlx5 real-time CQ timestamps only when `mlx5dv_query_device()` advertises `MLX5DV_CONTEXT_FLAGS_REAL_TIME_TS`.
- Retain the default device-clock CQ format on unsupported devices and convert those timestamps through the existing `mlx5dv_ts_to_ns()` path.
- Preserve the valid #281 RX WQ UMEM access flags, adapter-relative page-size encoding, and reduced 64-stride region geometry.

## Failure and expected behavior

After #281, raw ibverbs RX fails inside `daqiri_init()` before any traffic measurement:

```text
CREATE_RQ (DevX) failed: Remote I/O error (syndrome 0x43d98f)
```

The identical configuration initializes with a pre-#281 image. Raw RX should also initialize on devices whose firmware does not support real-time CQ timestamps; those devices should use the free-running device clock and DAQIRI's existing clock conversion.

## Root cause and isolation

#281 unconditionally set `MLX5_RQC_TIMESTAMP_FORMAT_REAL_TIME` in the DevX RQ context. The affected device does not advertise the corresponding mlx5dv capability, and firmware rejects the unsupported RQ context during `CREATE_RQ`.

The RX-side changes were restored independently in diagnostic container builds:

| Diagnostic change | Result |
| --- | --- |
| Unmodified `main` | `CREATE_RQ` fails, syndrome `0x43d98f` |
| Restore pre-#281 512-stride region geometry | Same failure |
| Restore pre-#281 RX WQ UMEM access flags | Same failure |
| Restore both geometry and UMEM flags | Same failure |
| Omit only the unsupported real-time timestamp request | Initialization succeeds |
| Capability-gated timestamp request with #281 UMEM flags and 64-stride geometry | Initialization succeeds |

Increasing the RX buffer count to restore the old effective queue depth likewise does not change the failure.

The old setup succeeds because it leaves the RQ in the default free-running timestamp format. The corrected setup makes the same choice only when required by device capability; capable devices retain the real-time format and its direct UTC decode.

`log_wq_pg_sz` remains encoded relative to the mlx5 adapter's fixed 4-KiB page, rather than the operating-system page used to align and register the UMEM. This is why the adapter-relative value remains valid on ARM64 kernels using larger pages. The currently reachable validation systems report 4-KiB kernel pages, so a live 64-KiB-kernel confirmation remains explicitly outstanding while this PR is a draft.

### Build

The final commit was built and installed successfully with the project container:

```bash
IMAGE_TAG=daqiri:ibverbs-rx-init-fix-final \
BASE_TARGET=dpdk \
DAQIRI_ENGINE="dpdk ibverbs" \
scripts/build-container.sh
```

`git-clang-format --diff origin/main` reports no changes, and `git diff --check` passes.

### Initialization correctness proof

An RX-only raw ibverbs run on the affected hardware created the CQ and 64-stride DevX RQ, installed the flow, started the worker, and exited cleanly after five seconds:

```text
Striding RQ: stride=8192B strides/WQE=64 region=524288B num_wqe=512
real-time CQ timestamps are unsupported; using the device clock format
DevX striding RQ ready
ibverbs backend initialized with 1 RX queue(s), 0 TX queue(s)
RX complete: packets=0 bytes=0 bursts=0
```

The zero-packet RX-only run proves initialization only; it is not a throughput result.

### Physical cross-host completion and throughput proof

A subsequent physical cross-host raw ibverbs TX/RX run used the final fixed image on RX and an unchanged existing ibverbs image on TX. No benchmark configuration or harness semantics were changed.

- Active TX interval: 10.0131 seconds
- Application TX: 14,860,288 packets, 119,833,362,432 bytes
- Application RX: 14,860,288 packets, 119,833,362,432 bytes
- Application payload throughput over the active interval: 95.741 Gb/s TX and RX
- PHY TX delta: 14,860,292 packets, 119,892,804,548 bytes
- PHY RX delta: 14,860,292 packets, 119,892,804,548 bytes
- The four extra PHY packets and 964 bytes are background control traffic; the data-packet PHY byte count is exactly the application packet count multiplied by the 8,068-byte wire frame.
- Stable one-second physical samples were approximately 97.0-97.4 Gb/s TX and 96.6-97.3 Gb/s RX.
- RX `out_of_buffer` remained flat; CQ errors and application-ring-full drops were zero.

This physical run proves that the corrected RX path receives and completes cross-host traffic at line-rate-class load. It is separate from the initialization proof and does not update or replace the published performance tables.

## Environment

- ARM64 NVIDIA DGX Spark
- Ubuntu 24.04.4 LTS
- Kernel `6.17.0-1014-nvidia`
- NVIDIA ConnectX-7 MT2910 family (`15b3:1021`)
- Firmware `28.45.4028`
- rdma-core/libibverbs/providers `2407mlnx52-1.2407061`

## Regression coverage

The failure occurs in device firmware during DevX object creation and there is no hardware-backed unit-test framework in the repository. The initialization and physical cross-host procedures above provide the regression check for this draft. A live run on a 64-KiB ARM64 kernel remains the only outstanding validation item.

## Scope

This PR changes only the ibverbs raw-RX timestamp-format selection and matching decode path. It does not alter TCP/socket behavior, benchmark documentation, published performance tables, or cross-host harness semantics.

### DIFF
--- a/src/engines/ibverbs/daqiri_ibverbs_engine.cpp
+++ b/src/engines/ibverbs/daqiri_ibverbs_engine.cpp
@@ -1005,7 +1005,7 @@ Status IbverbsEngine::devx_create_rq(IbvRxQueue& q, uint32_t stride_log, uint32_
     // the public PTP epoch-nanosecond timestamp contract.
     DEVX_SET(rqc, rqc, ts_format, MLX5_RQC_TIMESTAMP_FORMAT_REAL_TIME);
   } else {
-    DAQIRI_LOG_WARN(
+    DAQIRI_LOG_CRITICAL(
         "RX queue {}: real-time CQ timestamps are unsupported; using the device clock format",
         q.queue_id);
   }
--- a/src/engines/ibverbs/daqiri_ibverbs_engine.cpp
+++ b/src/engines/ibverbs/daqiri_ibverbs_engine.cpp
@@ -997,17 +997,25 @@ Status IbverbsEngine::devx_create_rq(IbvRxQueue& q, uint32_t stride_log, uint32_
   DEVX_SET(rqc, rqc, cqn, q.dv_cq.cqn);
   DEVX_SET(rqc, rqc, flush_in_error_en, 1);
   DEVX_SET(rqc, rqc, vsd, 1);  // do not strip VLAN
-  // The public timestamp contract is PTP epoch nanoseconds. Request mlx5
-  // real-time CQEs, whose timestamp field uses UTC encoding
-  // (seconds << 32 | nanoseconds).
-  DEVX_SET(rqc, rqc, ts_format, MLX5_RQC_TIMESTAMP_FORMAT_REAL_TIME);
+  struct mlx5dv_context dv_ctx {};
+  q.realtime_timestamps = mlx5dv_query_device(q.ctx, &dv_ctx) == 0 &&
+                          (dv_ctx.flags & MLX5DV_CONTEXT_FLAGS_REAL_TIME_TS) != 0;
+  if (q.realtime_timestamps) {
+    // Real-time CQEs use UTC encoding (seconds << 32 | nanoseconds), matching
+    // the public PTP epoch-nanosecond timestamp contract.
+    DEVX_SET(rqc, rqc, ts_format, MLX5_RQC_TIMESTAMP_FORMAT_REAL_TIME);
+  } else {
+    DAQIRI_LOG_WARN(
+        "RX queue {}: real-time CQ timestamps are unsupported; using the device clock format",
+        q.queue_id);
+  }
 
   DEVX_SET(wq, wq, wq_type, q.striding ? MLX5_WQ_TYPE_CYCLIC_STRIDING_RQ : MLX5_WQ_TYPE_CYCLIC);
   DEVX_SET(wq, wq, log_wq_stride, log2_floor(q.wqe_stride));
   DEVX_SET(wq, wq, log_wq_sz, log2_floor(q.num_wqe));
   DEVX_SET(wq, wq, pd, dvpd.pdn);
-  // WQ page size is encoded relative to the mlx5 4 KiB adapter page, not the
-  // operating-system page size (which is 64 KiB on GH200).
+  // The PRM field is relative to the mlx5 4 KiB adapter page, independent of
+  // the operating-system page size used to register the UMEM.
   DEVX_SET(wq, wq, log_wq_pg_sz, MLX5_ADAPTER_PAGE_SIZE_4_KIB);
   DEVX_SET64(wq, wq, dbr_addr, dbr_off);
   DEVX_SET(wq, wq, dbr_umem_id, q.wq_umem->umem_id);
@@ -3748,6 +3756,10 @@ Status IbverbsEngine::get_packet_rx_timestamp(BurstParams* burst, int idx, uint6
     return Status::INVALID_PARAMETER;
   }
   const uint64_t raw_timestamp = burst_ts_arr(burst)[idx];
+  if (!q->realtime_timestamps) {
+    *timestamp_ns = ts_to_ns(q->ctx, raw_timestamp);
+    return Status::SUCCESS;
+  }
   // mlx5 real-time CQEs use UTC encoding; expose the public API's linear PTP
   // epoch-nanosecond representation.
   const uint64_t seconds = raw_timestamp >> 32;
--- a/docs/api-reference/configuration.md
+++ b/docs/api-reference/configuration.md
@@ -349,6 +349,10 @@ a template table.
 When enabled, DAQIRI requires hardware timestamp support from the NIC and driver.
 Timestamps returned by `get_packet_rx_timestamp()` are unsigned 64-bit PTP epoch
 nanoseconds in the same clock domain as a PTP-synchronized `CLOCK_REALTIME`.
+The raw ibverbs engine requests the mlx5 real-time CQ timestamp format only when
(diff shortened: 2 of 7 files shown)


src/engines/ibverbs/daqiri_ibverbs_engine.cpp
```diff
@@ -997,17 +997,25 @@ Status IbverbsEngine::devx_create_rq(IbvRxQueue& q, uint32_t stride_log, uint32_
...
-  DEVX_SET(rqc, rqc, ts_format, MLX5_RQC_TIMESTAMP_FORMAT_REAL_TIME);
+  struct mlx5dv_context dv_ctx {};
+  q.realtime_timestamps = mlx5dv_query_device(q.ctx, &dv_ctx) == 0 &&
+                          (dv_ctx.flags & MLX5DV_CONTEXT_FLAGS_REAL_TIME_TS) != 0;
+  if (q.realtime_timestamps) {
+    // Real-time CQEs use UTC encoding (seconds << 32 | nanoseconds), matching
+    // the public PTP epoch-nanosecond timestamp contract.
+    DEVX_SET(rqc, rqc, ts_format, MLX5_RQC_TIMESTAMP_FORMAT_REAL_TIME);
+  } else {
+    DAQIRI_LOG_WARN(
```
cliffburdick: I would make this critical since we don't want to support it. We should find out why this is failing on Spark, though.

cliffburdick (approved):

src/engines/ibverbs/daqiri_ibverbs_engine.cpp
dleshchev: Done in `959ec9e` (`#283 - Raise ibverbs timestamp fallback log level`). I changed the unsupported real-time CQ timestamp fallback from `DAQIRI_LOG_WARN` to `DAQIRI_LOG_CRITICAL` and re-ran the IGX hardware-loopback validation; it now emits the critical log, creates the DevX striding RQ, transfers traffic, and exits with `cqe_errors=0` / `app_ring_full_drops=0`.

PR #286, opened 2026-09-01
#285 - Add direct polling latency benchmark and BlueFlame posting

cliffburdick:
Reduce the latency on TX direct poll mode by writing directly to BF registers

### DIFF
--- a/docs/benchmarks/raw_benchmarking.md
+++ b/docs/benchmarks/raw_benchmarking.md
@@ -268,14 +268,23 @@ cable/peer-port return with the mlx5 hardware self-loopback path.
 latency thread is running under `SCHED_FIFO` at priority `N`. It requires the corresponding
 realtime scheduling and memory-lock privileges; omit it to retain normal `SCHED_OTHER` behavior.
 
-The summary and optional per-sample CSV use these boundaries:
+For each packet, the measured path follows this timeline:
+
+```text
+t0  Application timestamps immediately before send_tx_burst()
+t1  send_tx_burst() returns
+t2  NIC records the packet's RX hardware timestamp
+t3  get_rx_burst() returns the matching packet to the application
+```
+
+The summary and optional per-sample CSV combine those points into these boundaries:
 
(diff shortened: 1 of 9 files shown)


docs/benchmarks/raw_benchmarking.md
```diff
@@ -238,6 +238,54 @@ After having modified the configuration file, ensure you have connected an SFP c
...
+For a single-port internal comparison, merge the TX and RX queues under one interface, set both
+`bench_latency` interface names to it, use that port's MAC and PTP device, and set
+`daqiri.cfg.loopback: "hw"`. This retains the same timing boundaries while replacing the physical
+cable/peer-port return with the mlx5 hardware self-loopback path.
+
+`--realtime-priority N` locks current and future mappings with `mlockall()` and verifies that the
+latency thread is running under `SCHED_FIFO` at priority `N`. It requires the corresponding
+realtime scheduling and memory-lock privileges; omit it to retain normal `SCHED_OTHER` behavior.
+
+The summary and optional per-sample CSV use these boundaries:
```
RamyaGuru: Maybe starting with the timeline of steps in the application will help to explain each latency metric? Something like this:

`t0` timestamp before `send_tx_burst()`  
`t1` `send_tx_burst()` returns  
`t2` RX NIC hardware timestamp  
`t3` `get_rx_burst()` returns matching packet

```text
tx_call_to_return_ns  [t0 -> t1]
tx_call_to_rx_hw_ns   [t0 -> t2]
rx_hw_to_app_ns       [t2 -> t3]
tx_call_to_app_ns     [t0 -> t3]

RamyaGuru (approved):

docs/benchmarks/raw_benchmarking.md
cliffburdick: Addressed in 8208e80. The documentation now introduces the per-packet t0–t3 timeline before the metric table, then maps each reported latency directly to its corresponding timestamp interval.

PR #311, opened 2026-09-16
#310 - Pack raw ibverbs huge regions into shared arenas

### LINKED ISSUE #310: [FEA] Use a hugepage memory allocator to handle many small allocations
**Is your feature request related to a problem? Please describe.**
On a system with only 1 GiB hugepages, DAQIRI currently will use an entire hugepage for each allocation when configured for memory type 'huge' and fall back regular pages if hugepages aren't available.  For example on a system with 2 rx queues, 22 tx queues, and 16 hugepages, all 16x 1 GiB hugepages are allocated and 8 regular page backed allocations are used.

Use of regular pages results in high latency in the tails of the latency distribution, due to infrequently but slow page swapping.

**Describe the solution you'd like**
Use a memory allocator to associate many small allocations to a single hugepage. 

**Describe alternatives you've considered**
None

**Additional context**
None

tmartin-gh:
## Summary

Aggregate same-NUMA DAQIRI-owned huge memory regions into the smallest compatible hugetlb arenas while keeping each logical MR independently registered. Treat kind: huge as a strict hugetlb request and fail initialization instead of falling back to regular pages.  Documentation is also updated.  Fixes #310 

## Testing

Tested on an NVIDIA MT43244 BlueField-3 integrated ConnectX-7 NIC.

- Built the DAQIRI ibverbs engine for both Arm and x86 targets.
- Exercised scheduled packet transmission with custom application linked with DAQIRI.
- Confirmed zero TX drops, skipped bursts, late builds, and late enqueues during the validated scheduled-TX runs.
- Confirmed hugepages were used, no regular page fallback


src/engine.cpp
```diff
@@ -493,62 +516,333 @@ static inline size_t align_ceil(size_t value, size_t align) {
...
+    // delivering SIGBUS when the requested node cannot satisfy the policy.
+    if (madvise(ptr, candidate.mapped_size, MADV_POPULATE_WRITE) != 0) {
+      const int error = errno;
+      munmap(ptr, candidate.mapped_size);
+      DAQIRI_LOG_WARN(
+          "Could not populate {} byte hugetlb arena using {} byte pages on NUMA node {}: {}",
+          candidate.mapped_size, candidate.page_size, numa, std::strerror(error));
+      continue;
+    }
+#endif
```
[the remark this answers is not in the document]
tmartin-gh: This appears to be a pre-Linux-Kernel-5.14 runtime-backport issue. Affected hosts might include:
* Ubuntu 20.04 GA with Linux 5.4
* RHEL/Rocky/Alma 8 with a 4.18-based kernel, unless the vendor backported it
* Any older host running a container built with newer Linux/glibc headers

DAQIRI documentation already recommended Ubuntu 22.04 which shipped initially with linux kernel 5.15.  Therefore, we have updated the documentation to reflect DAQIRI supports linux kernel 5.15+ instead of 5.4+.  We will not implement the compatibility case for older kernels that don't support `MADV_POPULATE_WRITE`

tmartin-gh:
@greptile full review

docs/api-reference/cpp.md
```diff
@@ -51,6 +51,11 @@ auto status = daqiri::daqiri_init(config);
 After `daqiri_init()` returns `Status::SUCCESS`, all memory regions are allocated, NIC
 queues are configured, and worker threads are running.
 
+`MemoryKind::HUGE` is an explicit hugetlb request for DAQIRI-owned memory. DAQIRI does not
+substitute regular or transparent-hugepage memory when the requested hugetlb allocation is
+unavailable; initialization fails instead. Use `MemoryKind::HOST` when regular host memory is
```
cliffburdick: or HOST_PINNED

tmartin-gh: Updated.

cliffburdick (approved):

